Issue #3, opened 2015-07-08
completion request should flush any unsyncd overlay changes


devoncarew:
We should add a `flush()` method to buffer updaters, so that a completion request operates on the very latest changes.

radicaled:
This is a side-effect of switching to using onDidStopChanging instead of onDidChange inside of the buffer updaters -- so, now that we're not sending realtime changes to the analysis server, we have to manually synchronize whenever autocomplete is called. `dart-tools` takes this approach and in retrospect it is a pretty bad one: we'll have to manually sync every editor pane before every action. Formatting, quick fixes, autocomplete, etc.

It leaves a lot of room for future errors of the data inconsistency kind, I think, and makes the system harder to reason about.

devoncarew:
I agree that changing back to `onDidChange` from `onDidStopChanging` will make for a simpler system, with less room for programmer error. I'm worried about the UX of immediately telling users about warnings. They should have some grace period of typing where we don't tell them if there are issues with their code. Just delaying reporting the results by a fixed period will just push the problem down the road a second or two.

We could introduce some buffing logic in the `ErrorRepository` class. The right logic might mean that we can use the `onDidChange` event w/o any negative UX.

radicaled:
In my opinion the correct approach is having the underlying system be as real time as possible while its consumers do whatever it is they need to do to handle the incoming data. `ErrorRepository` would be a bad place to do the flow control in that respect. What if a non-ui component wanted real time access to the errors data?

I think that'd be better handled at the consumer level: `DartLinterConsumer` should determine what batch of errors is most appropriate to show, and when. 

NB: If `Timer.periodic` worked in Atom we could just use the baked-in `debounce` method on the error stream `DartLinterConsumer` users.

devoncarew:
There's an issue with `Timer.periodic`?

radicaled:
I thought I emailed about this, but I can't find it so I guess not:

I can't get Timer.periodic to work in Atom -- it immediately generates some kind of exception from the bowels of dart2js. `debounce` uses `Timer.periodic` under the hood. A normal timer seems to work fine.

devoncarew:
> In my opinion the correct approach is having the underlying system be as real time as possible while its consumers do whatever it is they need to do to handle the incoming data. ErrorRepository would be a bad place to do the flow control in that respect. What if a non-ui component wanted real time access to the errors data?

[...]

PR #24, opened 2015-07-09
use onDidChange instead

devoncarew:
Fix https://github.com/dart-atom/dartlang/issues/3 - revert back to using `onDidChange`.

@radicaled


radicaled:
Looks good to me. If `Timer.periodic` is resolved, the delay can be handled with `debounce`: http://www.dartdocs.org/documentation/frappe/0.4.0%2B4/index.html#frappe/frappe.EventStream@id_debounce

Should have the same behavior, semantically, as using onDidChange, just at a different point in the system.

PR #46, opened 2015-07-11
Add an option to filter warnings about compling to JS

abarth:
Some Dart projects won't be compiled to JavaScript, which means these warnings
are just spam. For example, projects that depend on dart:sky won't be compiled
to JavaScript because dart:sky cannot be compiled to JavaScript.

Example warning: "When compiled to JS, this test might return true when the left hand side is an int"

### DIFF
--- a/lib/plugin.dart
+++ b/lib/plugin.dart
@@ -153,9 +153,15 @@ class AtomDartPackage extends AtomPackage {
         'type': 'boolean',
         'default': false
       },
-      'showUnnamedLibraryWarning': {
+      'filterUnnamedLibraryWarnings': {
         'title': 'Filter unnamed library warnings',
-        'description': 'Filter all unnamed library warnings.',
+        'description': 'Don\'t display warnings about unnamed libraries.',
+        'type': 'boolean',
+        'default': true
+      },
+      'filterCompiledToJSWarnings': {
+        'title': 'Filter warnings about compiling to JavaScript',
+        'description': 'Don\'t display warnings about compiling to JavaScript.',
         'type': 'boolean',
         'default': true
       }
--- a/lib/linter.dart
+++ b/lib/linter.dart
@@ -53,8 +53,10 @@ final Map<String, String> _severityMap = {
 
 String _infosPrefPath = '${pluginId}.showInfos';
 String _todosPrefPath = '${pluginId}.showTodos';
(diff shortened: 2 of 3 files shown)


radicaled:
@devoncarew I'm starting to feel like we need a `.dart-project`  file or something -- a lot of these settings are going to be project-specific, esp. if you bounce across multiple Dart projects.

lib/plugin.dart
```diff
@@ -153,9 +153,15 @@ class AtomDartPackage extends AtomPackage {
         'type': 'boolean',
         'default': false
       },
-      'showUnnamedLibraryWarning': {
+      'filterUnnamedLibraryWarnings': {
```
devoncarew: While slightly inaccurate, this setting was named `showUnnamed...` in order for it to display after the other settings in the preferences page. They're sorted (I believe) by pref id. `filter` will sort before everything else, and these aren't the primary prefs for the app.

We can play with some id prefixes in order to group settings; let's keep it as you have for now.

devoncarew:
@abarth, lgtm! What's an example of a `When compiled to JS` warning message? Can you add one to this PR's notes?

@radicaled, yup, agreed. We're making these settings global, when a user might really want some on for certain projects and off for others. The setting that @abarth contributed will be something that you'd likely want on for that project for any developer of it, which does feel like a setting you'd commit to the project repo.

@bwilkerson @pq have talked about a project level analysis settings file, which the analyzer and analysis server would use for things like configuring lints. That sounds similar to what we'd want. Not sure what the timeframe for that is, or if the intention is for it to also be able to configure spec warnings.

And re: the existing filter - for unnamed library warnings - the latest on that is that the spec will change to remove the warning. So that particular filter is short-lived.

abarth:
> What's an example of a When compiled to JS warning message?

The one I'm hitting is "When compiled to JS, this test might return true when the left hand side is an int".

pq:
> @bwilkerson @pq have talked about a project level analysis settings file, which the analyzer and analysis server would use for things like configuring lints. That sounds similar to what we'd want. Not sure what the timeframe for that is, or if the intention is for it to also be able to configure spec warnings.

Shouldn't be too far off.  It occurs to me though that we should track this with an issue...  Could you [open one](https://github.com/dart-lang/sdk/labels/Area-Analyzer) and assign it me?  Thanks!

Issue #49, opened 2015-07-12
have a task to bump our version number for releases


devoncarew:
The task should bump the version number and keep the pubspec, package.json, and changelog in sync (https://github.com/dart-atom/dartlang/pull/41#issuecomment-120557840).

PR #50, opened 2015-07-12
Automate release tagging for package.

radicaled:
Majority of the work for #49.

@devoncarew 

I normally use Rake, so my knowledge of Grinder is pretty limited. The logic works fine, but it's hard-coded to "nextPatch" instead of accepting a command line parameter. You'll probably want to tweak the changelog pattern / output text, too.

### DIFF
--- a/tool/deploy.dart
+++ b/tool/deploy.dart
@@ -1,0 +1,60 @@
+import 'dart:io';
+
+import 'package:grinder/grinder.dart';
+import 'package:pub_semver/pub_semver.dart';
+
+main(List args) => grind(args);
+
+@DefaultTask('Deploy software')
+deploy() {
+  deployIt();
+}
+
+deployIt() {
+  // TODO: command line parsing (major, minor, patch) version update
+  // Comment out the if statements while developing this task.
+  var diffResult = Process.runSync("git", ["diff", "--shortstat"]);
+  if (diffResult.stdout.toString().isNotEmpty)
+    throw 'Index is dirty; commit files and try again';
+
+  var syncResult = Process.runSync("git", ["rev-list", "origin..HEAD"]);
+  if (syncResult.stdout.toString().isNotEmpty)
+    throw 'Not synchronized with master; git push and try again';
+
+  var packageFileData = new File('package.json').readAsStringSync();
+  var yamlFileData    = new File('pubspec.yaml').readAsStringSync();
+  var changelogData   = new File('CHANGELOG.md').readAsStringSync();
+
+  var packageData = JSON.decode(packageFileData);
+
+  var currentVersion = new Version.parse(packageData['version']);
+  // TODO: have this match the parameter passed in on the command line
+  var nextVersion = currentVersion.nextPatch;
+
+  var jsonPattern         = (version) => '"version": "${version}"';
+  var yamlPattern         = (version) => 'version: ${version}';
+  var changelogPattern    = (version) => "# ${version}"
+  // We do pattern matches to preserve the formatting in the existing files.
+
+  var newPackageFileData =
+    packageFileData.replaceFirst(jsonPattern(currentVersion),
+      jsonPattern(nextVersion));
+  var newYamlFileData =
+    yamlFileData.replaceFirst(yamlPattern(currentVersion),
+      yamlPattern(nextVersion));
+  var newChangelogData =
+    changelogData.replaceFirst(changelogPattern('Upcoming Version'),
+      changelogPattern(nextVersion));
+
+  new File('package.json').writeAsStringSync(newPackageFileData);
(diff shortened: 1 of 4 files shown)


tool/deploy.dart
```diff
@@ -0,0 +1,60 @@
...
+
+main(List args) => grind(args);
+
+@DefaultTask('Deploy software')
+deploy() {
+  deployIt();
+}
+
+deployIt() {
+  // TODO: command line parsing (major, minor, patch) version update
```
devoncarew: `grinder` doesn't yet support passing args to tasks, but this is under active development.

tool/deploy.dart
```diff
@@ -0,0 +1,60 @@
...
+
+  var syncResult = Process.runSync("git", ["rev-list", "origin..HEAD"]);
+  if (syncResult.stdout.toString().isNotEmpty)
+    throw 'Not synchronized with master; git push and try again';
+
+  var packageFileData = new File('package.json').readAsStringSync();
+  var yamlFileData    = new File('pubspec.yaml').readAsStringSync();
+  var changelogData   = new File('CHANGELOG.md').readAsStringSync();
+
+  var packageData = JSON.decode(packageFileData);
```
devoncarew: Need to import `dart:convert`?

tool/deploy.dart
```diff
@@ -0,0 +1,60 @@
...
+
+  var currentVersion = new Version.parse(packageData['version']);
+  // TODO: have this match the parameter passed in on the command line
+  var nextVersion = currentVersion.nextPatch;
+
+  var jsonPattern         = (version) => '"version": "${version}"';
+  var yamlPattern         = (version) => 'version: ${version}';
+  var changelogPattern    = (version) => "# ${version}"
```
devoncarew: `;`

tool/deploy.dart
```diff
@@ -0,0 +1,60 @@
...
+  new File('pubspec.yaml').writeAsStringSync(newYamlFileData);
+  new File('CHANGELOG.md').writeAsStringSync(newChangelogData);
+
+  Process.runSync("git", ["commit", "-a", '-m "prepare $nextVersion"']);
+  Process.runSync("git", ["tag", nextVersion.toString()]);
+  Process.runSync("git", ["push", "-t"]);
+  Process.runSync("apm", ["publish", "-t", nextVersion.toString()]);
+
+  print("Version ${nextVersion} has been prepared!");
+  print('¯\\_(ツ)_/¯');
```
devoncarew: :)

tool/deploy.dart
```diff
@@ -0,0 +1,60 @@
+import 'dart:io';
+
+import 'package:grinder/grinder.dart';
+import 'package:pub_semver/pub_semver.dart';
+
+main(List args) => grind(args);
+
+@DefaultTask('Deploy software')
+deploy() {
```
devoncarew: You may want to just add this task to the main `tool/grind.dart` script instead of creating a new library. It can support multiple tasks, with or without dependencies on the other tasks. Run `grind -h` to see the list of defined tasks, and `grind <task>` to run a specific one.

devoncarew:
Very cool!

> The logic works fine, but it's hard-coded to "nextPatch" instead of accepting a command line parameter.

Yeah, grinder doesn't yet support passing cli args to individual tasks, but that's currently being developed.

radicaled:
Hm, I must have accidentally butchered this file before committing (that or too many CTRL-Zs by accident). I'll clean it up in a bit.

tool/deploy.dart
radicaled: Is it possible to expose new tasks to grinder from other Dart files? After a certain point navigating `tool/grind.dart` will become difficult it its a large file.

radicaled:
> Yeah, grinder doesn't yet support passing cli args to individual tasks, but that's currently being developed.

This PR is pretty isolated, so it should probably just wait around until that lands in grinder.

tool/deploy.dart
devoncarew: Yeah, just import the library that defines the task, that should do it. You _may_ have to export the library for the annotation scanner to see that task.

radicaled: That actually seemed to raise an exception, so I just made `publish.dart` (formerly `deploy.dart`) part of `atom.grind` instead.

devoncarew:
Here's the grinder issue: https://github.com/google/grinder.dart/issues/175.

devoncarew:
Hmm, just accidentally published a new version w/o first updating the changelog title from `## unpublished`. Perhaps we could land this and iterate on it when grinder args are a thing?

radicaled:
Sounds good to me.

PR #53, opened 2015-07-13
sort less used pref. settings to the bottom

devoncarew:
Prefix less used pref settings with `x_` in order to have them display at the end of our preferences dialog (things like the sdk location should be at the top).

before:
screen shot 2015-07-13 at 1 36 47 pm

after:
screen shot 2015-07-13 at 1 42 18 pm


devoncarew:
@abarth and @radicaled

devoncarew:
(the `Logging` setting will only show up for developers of the plugin)

PR #63, opened 2015-07-15
Improve some grammar highlighting

austincummings:
Makes function highlighting more consistent. Also removed `String` from primitives as that will be highlighted by the class highlighting rule.

### DIFF
--- a/grammars/dart.cson
+++ b/grammars/dart.cson
@@ -165,8 +165,10 @@
         'name': 'support.class.dart'
       }
       {
-        'match': '(_?[a-z][a-zA-Z]+)(?=\\\(\\\[)'
-        'name': 'entity.name.function.dart'
+        'match': '(_?[a-z][a-zA-Z0-9]+)\\('
+        'captures':
+          '1':
+            'name': 'entity.name.function.dart'
       }
       {
         'match': '\\b[A-Z_0-9]+\\b'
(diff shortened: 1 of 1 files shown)


radicaled:
Are strings in dart considered primitives? If so, we should leave that rule in, in case someone does a theme / other package wants to do something with primitives.

austincummings:
I don't think Dart has primitives in the traditional sense. Everything in Dart is an object, even `int`'s, `double`'s, and `bool`'s. The difference I see between `String` and `int` is that `String` is capitalized like  a class, and thus it should be treated like a class in my opinion.

radicaled:
> The difference I see between String and int is that String is capitalized like a class, and thus it should be treated like a class in my opinion.

@devoncarew what do you think about this? it means `String` will be considered/colorized exactly the same as classes for some (all?) syntax themes.

devoncarew:
I'm honestly on the fence about it. `String` is somewhere in between a 'primitive' type like `int` and `bool` and a user defined type. It's common and core to dart programing and used at about the same frequency as the primitive types.

On the other hand, none of `num`, `int`, `double`, or `bool` are extendable by the user. `String` _is_ extendable. And, it's the only one that uses title case, like regular user defined classes.

It may just come down to trying it one way and seeing if that feels odd. tldr, I'm ok either way.

radicaled:
I guess that's fine by me, unless someone on the Dart language team pipes in?

devoncarew:
Let's try it like this for a while; if it doesn't feel right after using it, we can always switch back :)

PR #84, opened 2015-07-18
Devoncarew dartdoc hover

devoncarew:
- clean up JS interop in navigation.dart
- refactor dartdoc related code into `dartdoc.dart`
- make a new modal dialog for dartdoc info; iterate on the styling

Dartdoc still shows up on `F1`. Having it show on mousehover involved way too many JS interop hoops. Perhaps if some atom plugin shows up that supports pluggable hover providers we can switch over to that.

screen shot 2015-07-18 at 2 10 06 am

screen shot 2015-07-18 at 2 10 21 am


radicaled:
> Having it show on mousehover involved way too many JS interop hoops. 

Could just write this part in JavaScript and then call it with a single JS Interop call?

devoncarew:
That would work :)

The typescript plugin has some tooltips code, as does a few others. I get the sense that the same code's being copied around.

The hard part was going from a mouse event's coordinates to an editor's buffer coordinates.

Issue #106, opened 2015-07-21
Uncaught TypeError: Cannot read property 'analyzing' of null


zanderso:
[Enter steps to reproduce below:]

1. Packages > Dart > Show Analysis Server Status
2. ...

**Atom Version**: 1.0.2
**System**: Ubuntu 14.04
**Thrown From**: [dartlang](https://github.com/dart-atom/dartlang) package, v0.3.4

### Stack Trace

Uncaught TypeError: Cannot read property 'analyzing' of null

```
At /usr/local/google/home/zra/.atom/packages/dartlang/web/entry.dart.js:117

TypeError: Cannot read property 'analyzing' of null
  at AnalysisServer.dart.AnalysisServer.get$isBusy (/usr/local/google/home/zra/.atom/packages/dartlang/web/entry.dart.js:16655:28)
  at AnalysisServerDialog.dart.AnalysisServerDialog._updateStatus$0 (/usr/local/google/home/zra/.atom/packages/dartlang/web/entry.dart.js:17281:67)
  at AnalysisServerDialog.dart.AnalysisServerDialog.showDialog$0 (/usr/local/google/home/zra/.atom/packages/dartlang/web/entry.dart.js:17267:14)
  at AnalysisServerDialog_closure.dart.AnalysisServerDialog_closure.call$1 (/usr/local/google/home/zra/.atom/packages/dartlang/web/entry.dart.js:17325:62)
  at CommandRegistry_add_closure.dart.CommandRegistry_add_closure.call$1 (/usr/local/google/home/zra/.atom/packages/dartlang/web/entry.dart.js:16284:48)
  at [object Object].dart.Primitives_applyFunctionWithPositionalArguments (/usr/local/google/home/zra/.atom/packages/dartlang/web/entry.dart.js:13602:28)
  at dart._callDartFunction (/usr/local/google/home/zra/.atom/packages/dartlang/web/entry.dart.js:28761:31)
  at atom-workspace.<anonymous> (/usr/local/google/home/zra/.atom/packages/dartlang/web/entry.dart.js:28749:18)
  at CommandRegistry.module.exports.CommandRegistry.handleCommandEvent (/usr/share/atom/resources/app.asar/src/command-registry.js:241:29)
  at CommandRegistry.handleCommandEvent (/usr/share/atom/resources/app.asar/src/command-registry.js:3:61)
[...]

Issue #77, opened 2015-07-16
Lost code completion, analysis server is idle


sethladd:
Not sure how to provide information here. I've lost code completion.

* No spinning gear for analysis server.
* Analysis server status says idle.
* Activity Monitor doesn't show any 100% CPU

If there's a log that I can provide, please let me know.

![screen shot 2015-07-16 at 1 12 09 pm](https://cloud.githubusercontent.com/assets/5479/8734069/4f2c56dc-2bbc-11e5-9768-b7864acb3817.png)

sethladd:
After a restart, now my analysis server seems to be analyzing everything in my .pub-cache, but very very slowly. The analysis gear is spinning.

This is a "real time" gif:

![2015-07-16 13_24_52](https://cloud.githubusercontent.com/assets/5479/8734311/1098fe82-2bbe-11e5-91b7-ada2c7d7bedb.gif)

sethladd:
Now I'm not seeing errors or warnings, and I see "No Issues", when I should see one error.

![screen shot 2015-07-17 at 9 55 49 am](https://cloud.githubusercontent.com/assets/5479/8752616/1b629c9a-2c6a-11e5-9472-1872faa12839.png)

PR #110, opened 2015-07-21
fix a performance issue with opening files

### LINKED ISSUE #105: opening dart:html causes the analysis server to melt down

devoncarew:
Fix #105. Atom sends a lot of change events when a file is opened. The number of events is proportional to the size of the file. For average file, this is 3-10 events. We introduce a very short frappe debounce to address this. We'll have to be aware that this _might_ interact poorly with code completion - we might be completing on a file w/o the last typed char.

For very large files, atom sends hundreds of events. For dart:html it was over 800 events. We put in a check to make sure we're not sending change overlays for files that are not in dart projects (like the sdk files).


devoncarew:
fix #106

devoncarew:
fixes https://github.com/dart-atom/dartlang/issues/77

Issue #103, opened 2015-07-21
Run dartfmt on save configuration option


halverneus:
Provide the user the configuration option to run dartfmt on save.

laike9m:
It seems I can't trigger auto format. I'm using Atom 1.9.9 with "Format files on save" checkbox selected.

PR #113, opened 2015-07-23
Add format on save

radicaled:
Some minor refactoring required.


lib/atom.dart
```diff
@@ -646,6 +646,9 @@ class TextBuffer extends ProxyHolder {
   /// the next line.
   Range rangeForRow(int row, bool includeNewline) =>
       new Range(invoke('rangeForRow', row, includeNewline));
+
+  /// Invoke the given callback before the buffer is saved to disk.
+  Stream get onWillSave => eventStream('onWillSave');
```
devoncarew: thinking out loud, it might be nice to have an optional arg to `eventStream` to make the broadcast stream synchronous.

lib/analysis/formatting.dart
```diff
@@ -15,22 +17,24 @@ final Logger _logger = new Logger('formatting');
 int get _prefLineLength =>
     atom.config.getValue('editor.preferredLineLength', scope: ['source.dart']);
 
+// TODO: this should be "FormattingCommandManger" (?), containing only logic
```
devoncarew: Feel free to rename :)

devoncarew:
+1, this will be great to have. I thought there was a related issue, but can't find one now...

radicaled:
It's probably #103

lib/analysis/formatting.dart
radicaled: too lazy. leaving it for the next sucker to go through that file.

Issue #117, opened 2015-07-23
Reanalyzing if analysis server was shutdown will generate a red error


sethladd:
Shutdown the analysis service.
Then, select "Reanalyze sources" from the Packages | Dart menu.

I get this error:

![screen shot 2015-07-23 at 4 39 20 pm](https://cloud.githubusercontent.com/assets/5479/8864667/8bc5600a-3159-11e5-862c-f1bd292d3cb6.png)

radicaled:
Fixed by #119.

sethladd:
Thanks!

Issue #109, opened 2015-07-21
warn when the `linter` package is not installed


devoncarew:
From https://github.com/dart-atom/dartlang/issues/106. And / or install it for the user (and notify them).

radicaled:
Fixed by #119

Issue #107, opened 2015-07-21
add a close button to the analysis server status window (was: Cannot dismiss analysis server status window)


zanderso:
Steps to repro:
1. Packages > Dart > Show Analysis Server Status
2. Try to make status window go away.

What happens:
Window does not go away.

What should happen:
Window should go away.

devoncarew:
You can dismiss it by hitting `escape`. That said, we should add a close button :)

zanderso:
Of course escape works for me only after I file the issue =) A little [x] would be good, yah.

devoncarew:
I'll re-purpose this issue -

radicaled:
@devoncarew 

Looks like #119 fixed this -- verify and close issue if so?

devoncarew:
Yup, fixed.

PR #119, opened 2015-07-25
fix restart exception; add close to dialog; misc fixes

devoncarew:
- warn when packages that we require are not installed
- fix an NPE from the `re-analyze sources` command
- added a close button to the jobs dialog and the analysis server dialog


devoncarew:
fix #117, fix #109, fix #107

PR #121, opened 2015-07-28
implement a type hierarchy view

devoncarew:
Implement a type hierarchy view.

- create a new `AtomView` class which represents a panel on the right hand side of atom
- implement a movable splitter control
- expose atom's tree control through `ListTreeBuilder`
- handle selection and double clicking in the type hierarchy view
- add keybindings and context menu bindings for the type hierarchy and `find references` commands


devoncarew:
screen shot 2015-07-27 at 10 54 02 pm

Issue #189, opened 2015-08-07
analysis server can consume extreme amounts of memory


devoncarew:
From a user report:

```
I have a dartanalyzer using 17GB of my memory as I write this.

I had one earlier using 34GB.

Possibly to do with a symlink loop?
```

The resolution here might be to add some server diagnostics commands to atom.

devoncarew:
We're trying to address this with better diagnostics on the atom side.

PR #190, opened 2015-08-07
enable AS diagnostics

devoncarew:
Some diagnostics work in support of #189.

screen shot 2015-08-07 at 10 32 21 am


devoncarew:
@sethladd, FYI

PR #199, opened 2015-08-09
add the ability to create a new sky project

devoncarew:
- fix an issue with quick fixes selecting the right char range in the wrong file
- improve the UI of the `promptUser` dialog
- add a `PubApp` class - it can install and run dart apps installed via pub
- add the ability to create a new sky project. This uses the `sky_tools` package, and it exposed via the `dartlang:create-sky-project` command and the Dart package menu


devoncarew:
screen shot 2015-08-09 at 2 16 41 pm

PR #222, opened 2015-08-14
Stop consuming CMD+1 on MacOS

johnmccutchan:


keymaps/atom-dart.cson
```diff
@@ -21,8 +21,8 @@
   'ctrl-shift-O': 'dartlang:organize-directives'
 
 # Dart editor commands (mac)
+# TODO: Add default keybinding for dartlang:quick-fix on MacOS.
```
devoncarew: Let's go with `ctrl-1` for all platforms. Can you remove the todo and move the ctrl-1 binding from below up to the `common to all platforms` section?

johnmccutchan:
Done.

devoncarew:
Thanks for the PR! Agree it's better to not step on atom default bindings when we can avoid it.

PR #223, opened 2015-08-14
add a context menu for running sky app

keertip:
@devoncarew 

start of a framework for launching and managing sky apps.


lib/sky/run_app.dart
```diff
@@ -0,0 +1,66 @@
...
+}
+
+class RunSkyAppContextCommand extends ContextMenuItem {
+  RunSkyAppContextCommand(String label, String command) : super(label, command);
+
+  bool shouldDisplay(AtomEvent event) {
+    _logger.info('showing sky menu');
+    String filePath = event.targetFilePath;
+    _logger.info('filepath = $filePath');
+    DartProject project = projectManager.getProjectFor(filePath);
```
devoncarew: we should check for a null project here

lib/sky/run_app.dart
```diff
@@ -0,0 +1,66 @@
...
+      return result.stdout;
+    });
+  }
+}
+
+class RunSkyAppContextCommand extends ContextMenuItem {
+  RunSkyAppContextCommand(String label, String command) : super(label, command);
+
+  bool shouldDisplay(AtomEvent event) {
+    _logger.info('showing sky menu');
```
devoncarew: we probably don't need logging here, when checking to see if a context menu item should be contributed.

devoncarew:
lgtm! two suggestions

lib/sky/run_app.dart
keertip: Missed that! removed.

lib/sky/run_app.dart
keertip: done

Issue #246, opened 2015-08-24
dartlang makes bogus suggestion after comma in list


eseidelGoogle:
I hit this when making lists.

Here is example code to reproduce:
```
import 'dart:sky';
import 'package:sky/theme/colors.dart' as colors;

List<Color> foo = [
  colors.white,
];
```

Type `colors.black,` on the next line.  It will attempt to then suggest `false` to follow the comma. When you hit enter to try to go to go to the next line, you'll end up adding `false` and being confused.

![image](https://cloud.githubusercontent.com/assets/11857803/9455167/3d4efe94-4a7d-11e5-8b7e-52c043973873.png)

@devoncarew

Issue #239, opened 2015-08-21
release notes popup appears twice


skybrian:
It seemed like I was seeing the release notes popup rather often, so I did an experiment.

I just upgraded to 0.3.17. After quitting Atom and starting it again, the release notes popup appeared. Then I immediately shut down atom and launched it again, and the same popup appeared again.

I shut it down again and restarted, and it didn't appear the third time.

(Since I use atom on two different machines, that means I'm probably seeing the same release notes popup four times.)

devoncarew:
I'm investigating, but ironically, 0.3.17 contains code to reduce the number of times we show the release notes. We had been storing the last version run on localStorage; for atom, it looks like that data store is based on the dir that atom was launched in.

We switched over to using a hidden pref in the user's config. That is a per-user setting. You can see it in your config file (`_dartlang: _version`).

Really surprised it showed up twice. When you say quit, did you quit atom, or was the plugin uninstalled and re-installed as part of the upgrade process? I ask because we do have an issue with some parts of the old plugin still being active after an upgrade, so the old code to show the changlog notes might have run.

skybrian:
Well, I thought I did. I don't have a way to confirm it now. I guess we can wait and see if it happens again.

skybrian:
I started Atom on my work computer (after not using it in a while) and it showed release notes for an old version of the Dart plugin. Then I went to Preferences-> Packages and clicked on the Upgrade button. It showed the release notes immediately (without rebooting). I then exited and started up, and it showed them again.

Perhaps that's what I did last time.

Issue #6, opened 2015-07-08
After a statement is completed the completions dialog gets in the way


sethladd:
(was https://github.com/dart-lang/atom-dartlang-experimental/issues/93)

type e.g. 

```
import 'dart:io' as io; 
```

Down arrow.

The completion dialog is still showing after the ';' which means that the down arrow, or an enter is swallowed by that dialog.

I'm going to wear out the esc key on my keyboard pretty quickly at this rate.

sethladd:
from @radicaled 

"Also, I highly recommend going into your package settings for autocomplete-plus and changing the confirmation token from "tab and enter" to just "tab." I don't know why (or when) they changed the default, but there are probably going to be times when autocomplete pops up and you don't need it and want to do a line return instead."

PR #247, opened 2015-08-25
work to improve completion

devoncarew:
- more work towards reducing code completion twitchiness
- disable completion after `;{},`
- disable completion as-you-type; the user can override this in the settings for `autocomplete-plus`
- improve the default scoring from analysis server for completion
- don't show the release notes at startup; they are now available from the `Packages > Dart > Release Notes` menu item


devoncarew:
fix https://github.com/dart-atom/dartlang/issues/246
fix https://github.com/dart-atom/dartlang/issues/239
fix #6

PR #268, opened 2015-09-01
only report usage for dart projects

devoncarew:


lib/usage.dart
```diff
@@ -79,12 +80,16 @@ void _activePaneItemChanged(_) {
...
-    _ga.sendScreenView('editor/${path.toLowerCase()}');
+
+  // Ensure that the file is Dart related.
+  if (isDartFile(path) || projectManager.getProjectFor(path) != null) {
+    int index = path.lastIndexOf('.');
+    if (index == -1) {
+      _ga.sendScreenView('editor');
+    } else {
+      path = path.substring(index + 1);
+      _ga.sendScreenView('editor/${path.toLowerCase()}');
```
radicaled: Is there any particular reason you're capturing the filename the user is opening in the editor?

In the end you will effectively have an entire list of a user's project directory if they open every file at least once.

devoncarew: This is the file type - we strip everything but the extension. So `editor/dart`, `editor/css`, ...

Totally agree that the full path would be too invasive.

radicaled: Gotcha, I read that logic backwards (thought you were capturing the path before the `.`). My bad.

devoncarew: I have an unrelated PR in progress - I'll create a local var here - `extension` - to clarify things a bit :)

Issue #137, opened 2015-07-30
We miss an Outline view for the current file


sethladd:
Is there a good way to get an outline view for the current file?

Not so much for navigating the file, but for understanding where we are, and for easy jumping around. Getting a visual perspective of location is quite helpful.

Thanks for considering!

sethladd:
cc @johnmccutchan

devoncarew:
Yeah, it's one of the biggest remaining things I miss. We have most of the necessary pieces in place - a lot of the code from the type hierarchy and find references views will be reusable.

polux:
I don't know atom well: would having an outline also automatically give us the equivalent of Eclispe's crtl+O (a.k.a. Dart editor's alt+O), that is a popup where you can filter and jump to members of the current file? If not, I'll open another feature request for it.

devoncarew:
That would be separate, but would share much of the same code. A separate feature request would be great.

polux:
Just created #216.

devoncarew:
In progress:

<img width="370" alt="screen shot 2015-08-28 at 2 38 37 pm" src="https://cloud.githubusercontent.com/assets/1269969/9557662/99f78432-4d92-11e5-9f9a-87c491c33fff.png">

It's functional, but I want to iterate a bit on the UI.

polux:
Exciting!

devoncarew:
<img width="246" alt="screen shot 2015-08-29 at 9 15 30 am" src="https://cloud.githubusercontent.com/assets/1269969/9562925/94037db4-4e2e-11e5-90ae-7a071fed1023.png">

sethladd:
Should we put the return type on the right? It might look better if aligned
on the name of the property/method.

devoncarew:
I experimented with that. I tried:
- return type on the right for functions (using a right arrow char to separate params and return type)
- return type on the right for fields/vars (w/ a right arrow, and also with just a space separator)
- icons on the left to indicate different types of elements (classes, getters, fields, functions, ...)

When developing, I found that you had to think about the reversed notation. You just get very used to reading the source one way. You can scan stuff presented in that way faster than in other ways. And the left aligned names, while nice, actually made a ragged left edge in the outline view. private and public names would appear to be indented at a slightly different level.

I'm going to sign out for a bit, but will post some screenshots of the alternatives. Happy to iterate on the UI -

PR #272, opened 2015-09-01
outline tweaks

devoncarew:
Final outline view tweaks. From using it, I'm preferring the types-on-the-left style, for familiarity, rather than the types-on-the-right (for visual alignment). YMMV

screen shot 2015-09-01 at 12 34 05 pm

screen shot 2015-09-01 at 12 39 10 pm


devoncarew:
fix https://github.com/dart-atom/dartlang/issues/137

devoncarew:
I also tried icons instead of `get`, `set`, `class`, but it's hard to find representative icons, and then the user would need to learn what each icons was meant to represent.

PR #323, opened 2015-09-16
Don't show custom errors pane if it is disabled in settings

agudulin:
It is quite annoying to see opened errors pane with "No issues" label for non-dart or even for dart projects on the first editor launch.


devoncarew:
thanks! lgtm

Issue #261, opened 2015-08-27
Unclear how to trigger a Quick Fix


sethladd:
For example:

![screen shot 2015-08-27 at 12 53 11 pm](https://cloud.githubusercontent.com/assets/5479/9531400/a0f9a9e6-4cba-11e5-8500-77cf95faeee6.png)

How do I easily solve this problem?

Strawman: Would expect something like a link/message that says "Quick Fix available", which is displayed near the warning itself.

devoncarew:
This is blocked on https://github.com/dart-lang/sdk/issues/23874. I'd love to be able to do this -

devoncarew:
done!

PR #389, opened 2015-10-05
expose a UI for quick fix notifications

devoncarew:
Initial UX for https://github.com/dart-atom/dartlang/issues/261:

screen shot 2015-10-05 at 1 33 22 pm

@sethladd @pq


devoncarew:
- clicking on the wrench will jump to the error in the file and show the quick-fix menu (which looks similar to the code completion menu)
- the first time the user clicks on the wrench (per atom configuration), we show a toast with information about the keybinding to use to see a quickfix menu

pq:
Wow.  Looks great!

PR #444, opened 2015-10-22
add tests for utils Property

devoncarew:
- add tests for the `lib/utils.dart` `Property` class


devoncarew:
(merging - fix is in another PR)

Issue #413, opened 2015-10-09
ensure that we're analyzing all the dart files we can


devoncarew:
Specifically, are we analyzing dart files in the pub_cache? In the SDK? Are there more services we can have for those files (outline view?) then we are currently showing?

PR #454, opened 2015-10-27
enable analysis features in more files

devoncarew:
fix https://github.com/dart-atom/dartlang/issues/413 - enable analysis features in more files


devoncarew:
screen shot 2015-10-26 at 11 47 57 pm

PR #456, opened 2015-10-27
fix ~40 strong mode issues

devoncarew:
- knock down ~40 strong mode issues


devoncarew:
@leafpetersen, more strong mode changes, fyi -

Issue #51, opened 2015-07-12
auto install atom package deps


devoncarew:
See https://github.com/travs/atom-package-dependencies. This would currently just be for `linter`.

devoncarew:
We warn when there's a dep we'd like but don't have. The atom team plans to add package deps as first class things.

PR #506, opened 2015-11-10
auto-install package deps

devoncarew:
Add the ability to auto-install package deps; depend on the `linter` package.


devoncarew:
fix https://github.com/dart-atom/dartlang/issues/51

Issue #458, opened 2015-10-27
Uncaught TypeError: Cannot read property 'endsWith$1' of undefined


lukechurch:
@devoncarew This happened once after upgrading the plugin. After restarting, it doesn't appear to to repeat, and I don't know what version I was on before I think it was 2 versions back.

I'm not sure there's much more info I can give you, sorry. Sending this more as an FYi. If it isn't useful, feel free to close the issue.

**Atom Version**: 1.0.19
**System**: Mac OS X 10.10.5
**Thrown From**: [dartlang](https://github.com/dart-atom/dartlang) package, v0.4.12

### Stack Trace

Uncaught TypeError: Cannot read property 'endsWith$1' of undefined

```
At /Users/lukechurch/.atom/packages/dartlang/web/entry.dart.js:21255

TypeError: Cannot read property 'endsWith$1' of undefined
    at [object Object].J.endsWith$1$s (/Users/lukechurch/.atom/packages/dartlang/web/entry.dart.js:58779:40)
    at BufferObserverManager__closure.dart.BufferObserverManager__closure.call$1 (/Users/lukechurch/.atom/packages/dartlang/web/entry.dart.js:27923:15)
    at Workspace_observeTextEditors_closure.dart.Workspace_observeTextEditors_closure.call$1 (/Users/lukechurch/.atom/packages/dartlang/web/entry.dart.js:24401:48)
    at [object Object].dart.Primitives_applyFunctionWithPositionalArguments (/Users/lukechurch/.atom/packages/dartlang/web/entry.dart.js:22226:28)
    at dart._callDartFunction (/Users/lukechurch/.atom/packages/dartlang/web/entry.dart.js:49341:31)
    at /Users/lukechurch/.atom/packages/dartlang/web/entry.dart.js:49602:20
    at Workspace.module.exports.Workspace.observeTextEditors (/Applications/Atom.app/Contents/Resources/app.asar/src/workspace.js:276:9)
    at JsObject.dart.JsObject.callMethod$2 (/Users/lukechurch/.atom/packages/dartlang/web/entry.dart.js:49452:44)
    at Workspace.dart.ProxyHolder.invoke$4 (/Users/lukechurch/.atom/packages/dartlang/web/entry.dart.js:31317:23)
    at Workspace.dart.ProxyHolder.invoke$2 (/Users/lukechurch/.atom/packages/dartlang/web/entry.dart.js:31326:21)
```

### Commands

```
```

### Config

```json
{
  "core": {
[...]

PR #507, opened 2015-11-10
fix npe

devoncarew:
Fix an NPE encountered when launching applications (fix https://github.com/dart-atom/dartlang/issues/458).


nalkon:
Thanks Devon!  

Is there a way to update my plugin to include this fix?

devoncarew:
We're pushing out a new version today - you should have the fix in a few hours :)

nalkon:
Great! Thanks!

PR #511, opened 2015-11-11
fix the options file highlighting

devoncarew:
@pq

- fixed the .analysis_options syntax file to not conflict with the yaml one

I experimented with registering two completers (one for dart and one for the options files). From the docs, you're supposed to be able to return an array of completers instead of just a single object. I didn't have success, but it's something we could pursue.


pq:
(And thanks!)

Issue #457, opened 2015-10-27
console view should hyperlink stacktraces


devoncarew:
<img width="913" alt="screen shot 2015-10-27 at 11 16 27 am" src="https://cloud.githubusercontent.com/assets/1269969/10768003/3cb7cdd4-7c9c-11e5-854a-3e47c872ff40.png">

sethladd:
Can we do this in a generic way, or do we have to do something special for Flutter?

devoncarew:
It can be generic - we'll recognize the stack trace line, and the launch type will handle resolving that to a source location.

sethladd:
Very cool.

sethladd:
Super cool, thanks!

PR #525, opened 2015-11-15
hyperlink exceptions in the console view to the corresponding source …

devoncarew:
Hyperlink exceptions in the console view to the corresponding source files.

- recognize potential stack frames in the console view
- use heuristics and the analysis server to resolve those references to the right files
- have the various launch types be responsible for knowing how to resolve a reference
- fix https://github.com/dart-atom/dartlang/issues/457


devoncarew:
screen shot 2015-11-14 at 10 16 31 pm

PR #530, opened 2015-11-16
add some sample launch files

devoncarew:
@sethladd - here are some possible launch files. They'll all need `path`s, so we know what files they apply to. And the launch `type` will tell us which launch handler to use.

The other info in there is for the use of the launch types, and will vary between types. The `pub` / `test` one is for a launch type we don't currently support - it's just in there to play with the format.

### DIFF
--- a/.atom/launches/tooltest.yaml
+++ b/.atom/launches/tooltest.yaml
@@ -1,0 +1,10 @@
+# Command-line launch configuration for tool/test.dart.
+type: cli
+path: tool/test.dart
+
+cli:
+  checked: true
+  args:
+    - foo
+    - bar
+    - baz
--- a/.atom/launches/utils_test.yaml
+++ b/.atom/launches/utils_test.yaml
@@ -1,0 +1,7 @@
+# Pub launch configuration for test/utils_test.dart.
+type: pub
+path: test/utils_test.dart
+
+pub:
+  app: test
+  args: -j1 --no-color test/utils_test.dart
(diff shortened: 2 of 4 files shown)


.atom/launches/utils_test.yaml
```diff
@@ -0,0 +1,7 @@
+# Pub launch configuration for test/utils_test.dart.
+type: pub
```
sethladd: does this mean `pub run` ? Maybe we want to specify that clearly.

.atom/launches/tooltest.yaml
```diff
@@ -0,0 +1,10 @@
+# Command-line launch configuration for tool/test.dart.
+type: cli
```
sethladd: Do we want to specify which versions of these types are compatible?

sethladd:
Very cool, thanks for sharing.

.atom/launches/utils_test.yaml
devoncarew: `pub_run`? makes sense to me

.atom/launches/tooltest.yaml
devoncarew: Do you mean versioning the file?

sethladd: Sorry, I meant version of the cli or flutter.

.atom/launches/utils_test.yaml
sethladd: sgtm

.atom/launches/tooltest.yaml
devoncarew: My sense is that that is orthogonal to this file.

Issue #524, opened 2015-11-13
add text file based configuration of app launches


devoncarew:
- [ ] persist created launch configurations in text files in the project
- [ ] make those files discoverable and easily user editable
- [ ] use the info in there when we next launch the same app

Issue #513, opened 2015-11-11
How do I run a Flutter app with checked mode / not checked mode?


sethladd:
I'm using Cmd-R which is great, but I don't know if checked mode is on or off, and I don't know how to configure it for a run.

Thanks for your help!

devoncarew:
We need to add support for this; this'll come as part of the work to have a UI to configure launch configurations.

sethladd:
If that UI is a long ways away, maybe a new keyboard shortcut? Cmd-Option-R ?

Issue #359, opened 2015-09-25
launch improvements


devoncarew:
Launch, version 2:
- [x] general launch framework, determine the right launch type handler to use for a given starting resource
- [x] launch type handler for CLI apps
- [x] for Flutter launch type, re-cycle the current terminal UI if possible
- [x] remove the Flutter specific text from the flutter launch commands; make them general launch commands
- [x] remember launch configurations
- [x] on ambiguity, run the last launch configuration for a project
- [x] show a UI for configuring a launch config
- [x] show a UI for selecting one of several possible launch configurations

devoncarew:
https://github.com/dart-atom/dartlang/pull/376

PR #531, opened 2015-11-17
Yaml config

devoncarew:
Store launch config information in yaml files. 

```
path: tool/test.dart
type: cli

cli:
  args: one two three
  checked: true
  debug: true
```


devoncarew:
fix https://github.com/dart-atom/dartlang/issues/524, fix https://github.com/dart-atom/dartlang/issues/513, fox https://github.com/dart-atom/dartlang/issues/359,

Issue #526, opened 2015-11-15
create a general view manager component


devoncarew:
- tabbed views
- if there's just one tab, then we don't show the tab chrome

Issue #394, opened 2015-10-05
UI nits for the run contexts/outputs


sethladd:
![screen shot 2015-10-05 at 2 52 40 pm](https://cloud.githubusercontent.com/assets/5479/10294723/f6b51f58-6b70-11e5-8d9c-dd62843eafa4.png)

I think I understand the bug to me "click me to debug this". Is that accurate?

Is the white square next to that supposed to me "stop"? If so, might I suggest using Red ? It took me a while to guess what the white square is. There's no hover-over hint, either.

It's not clear these are tabs, but I think they are supposed to be tabs? That is, select one to see the output of that thing.

I'm really concerned I have what appears to be four mobile apps running. Is that true? What's supposed to happen after I run a Flutter app repeatedly ?

Not clear if there's an "active" tab? Which tab means "the app running on your phone right now" ?

devoncarew:
Thanks for the feedback! There's lot to iterate on here.

The selected tab is the blue one. We could make it clearer that these are tabs. I didn't want to take up as much space as a normal tab would have.

We should be re-cycling flutter runs, so you normally won't see this many running applications.

Can add some tooltips to help users know what's going on.

sethladd:
recycling: yes, one tab per app

What does "flutter" mean, in the tab? How does it know that it's a flutter app? And shouldn't it be "Flutter" ?

Selecting different tabs doesn't change the app on my phone. Though it changes what tab is blue. But maybe this is solved when we start recycling tabs.

devoncarew:
Urk. It is changing the running app, they're all just showing the same text, and don't update the scroll position as you cycle between apps. Definitely needs UI work.

devoncarew:
- changed to re-cycle the flutter launch - you shouldn't see more than one flutter launch at a time now
- add some UI to make it clearer that the badges act as tabs (an underline beneath the tabs)
- save the scroll position when switching between tabs
- `flutter` is the type of launch. If run run a command-line app, you'll see `cli`
- tooltips added to help make it clearer what the button do
- I experimented with using red for the stop button - it didn't play well with the background color. We do have some constraints with the colors - we have to stay to a particular palette of colors, which themes can swap out. Will keep an eye out for usability improvements to the buttons

sethladd:
> you'll see cli

I haven't seen this in action yet, but should we consider "Dart CLI" ? I don't know if the "Dart" context will be obvious...

> I experimented with using red for the stop button - it didn't play well with the background color

[...]

PR #536, opened 2015-11-18
Viewgroup

devoncarew:
Re-write our view management to use real tabs.

- fix https://github.com/dart-atom/dartlang/issues/526
- use the tabs control from atom
- when there's just one tab in a group, don't show the tabs control, just the content
- when there are no tabs, hide the tabbed container completely


devoncarew:
fix https://github.com/dart-atom/dartlang/issues/435
fix https://github.com/dart-atom/dartlang/issues/394

PR #596, opened 2015-12-01
Update status view

devoncarew:
Remove existing status dialogs in favor of the status view:
- moved the 'show sdk status' dialog functionality into the new status view sidebar
- moved the 'analysis server status' dialog functionality into the new status view sidebar
- delete the old dialogs
- add a new strobe component in support of the status view
- improve the messaging when we try and auto-locate the sdk

@danrubel 

screen shot 2015-12-01 at 11 59 04 am


lib/sdk.dart
```diff
@@ -335,8 +324,10 @@ class SdkLocationJob extends Job {
 
   SdkLocationJob(this.sdkManager) : super('Auto locate SDK');
 
+  bool get quiet => true;
```
danrubel: Is this used anywhere?

danrubel:
LGTM

lib/sdk.dart
devoncarew: Yup, it's an override from the parent. It indicates that a 'success' toast should not be shown after the job finishes.

PR #602, opened 2015-12-02
improve the analysis server status view

devoncarew:
Improve the analysis server status view. Use the info from the `diagnostic.getDiagnostics` call.

screen shot 2015-12-02 at 11 00 08 am

@danrubel @pq


lib/js.dart
```diff
@@ -39,7 +41,18 @@ dynamic jsObjectToDart(JsObject obj) {
     String str = _browserJson.callMethod('stringify', [obj]);
     return JSON.decode(str);
   } catch (e, st) {
-    _logger.severe('toDartObjectViaWizardy', e, st);
+    _logger.severe('jsObjectToDart', e, st);
+  }
+}
+
+dynamic dartObjectToJS(dynamic obj) {
```
danrubel: Where is this used? Just trying to learn the system :)

danrubel:
LGTM

lib/js.dart
devoncarew: Ah, this and the `uncrackDart2js` method above I added when I was working to put a JS charting library into the analysis server view (http://www.chartjs.org/). We decided to pull it out to re-focus on giving the user more actionable info. I still want to keep the JS methods, since they were useful then and may be again :)

pq:
Late to the party but :+1: !

Issue #617, opened 2015-12-07
add the ability to pass arbitrary cli options to the analysis server


devoncarew:
@bwilkerson

PR #618, opened 2015-12-07
add the ability to specify arbitrary cli options to the analysis server

devoncarew:
- add the ability to specify arbitrary cli options to the analysis server
- fix https://github.com/dart-atom/dartlang/issues/617

@danrubel @bwilkerson


danrubel:
LGTM

lib/analysis_server.dart
```diff
@@ -546,14 +547,27 @@ class _AnalysisServerWrapper extends Server {
...
+    if (AnalysisServer.startWithDiagnostics) {
       arguments.add('--port=${AnalysisServer.DIAGNOSTICS_PORT}');
       _logger.info('analysis server diagnostics available at '
           '${AnalysisServer.diagnosticsUrl}.');
     }
 
+    // Start in checked mode?
     if (AnalysisServer.useChecked) {
-      arguments.insert(0, '--checked');
+      arguments.add('--checked');
```
bwilkerson: I think this change is wrong. The `--checked` is a VM argument, so it really does need to come before the other arguments. (Similar issue in entry.dart.js.)

bwilkerson:
LGTM

lib/analysis_server.dart
devoncarew: Thanks; updated.

devoncarew:
In order to pass arbitrary options to the analysis server, edit your `config.cson` file (Atom > Open your Config`), and in the `dartlang` section, add a value for `analysisServerOptions`. This can either be a string with space separated values for separate options, or a cson array `["--foo", "--bar"]`.

Issue #611, opened 2015-12-03
Uncaught NullError: method not found: 'invoke$1' on null


shieldwed:
1. Remove Dart project folder from the projects panel.

**Atom Version**: 1.2.4
**System**: Microsoft Windows 10 Pro
**Thrown From**: [dartlang](https://github.com/dart-atom/dartlang) package, v0.4.16

### Stack Trace

Uncaught NullError: method not found: 'invoke$1' on null

```
At C:\Users\shiel\.atom\packages\dartlang\web\entry.dart.js:21314

TypeError: Cannot read property 'invoke$1' of null
    at ProxyHolder_eventStream_closure.dart.ProxyHolder_eventStream_closure.call$0 (C:\Users\shiel\.atom\packages\dartlang\web\entry.dart.js:31827:38)
    at [object Object].dart._runGuarded (C:\Users\shiel\.atom\packages\dartlang\web\entry.dart.js:39406:38)
    at _AsyncBroadcastStreamController.dart._BroadcastStreamController._callOnCancel$0 (C:\Users\shiel\.atom\packages\dartlang\web\entry.dart.js:39802:11)
    at _AsyncBroadcastStreamController.dart._BroadcastStreamController._recordCancel$1 (C:\Users\shiel\.atom\packages\dartlang\web\entry.dart.js:39713:18)
    at _BroadcastSubscription.dart._ControllerSubscription._onCancel$0 (C:\Users\shiel\.atom\packages\dartlang\web\entry.dart.js:40965:46)
    at _BroadcastSubscription.dart._BufferingStreamSubscription._cancel$0 (C:\Users\shiel\.atom\packages\dartlang\web\entry.dart.js:41046:35)
    at _BroadcastSubscription.dart._BufferingStreamSubscription.cancel$0 (C:\Users\shiel\.atom\packages\dartlang\web\entry.dart.js:41030:14)
    at _ProjectConfigurations.dart._ProjectConfigurations.dispose$0 (C:\Users\shiel\.atom\packages\dartlang\web\entry.dart.js:32786:35)
    at LaunchConfigurationManager.dart.LaunchConfigurationManager._handleProjectRemoved$1 (C:\Users\shiel\.atom\packages\dartlang\web\entry.dart.js:32632:18)
    at [object Object].<anonymous> (C:\Users\shiel\.atom\packages\dartlang\web\entry.dart.js:22584:32)
[...]

PR #619, opened 2015-12-07
fix an npe when removing a dart project

devoncarew:
- fix an npe when removing a dart project. The first time a launch was created for a project, the `.atom/launch` directory doesn't exist. The file watching code was failing in these situations, and when later removing the dart project, we got an NPE from a dispose method.
- fix https://github.com/dart-atom/dartlang/issues/611

@danrubel


danrubel:
LGTM

PR #623, opened 2015-12-08
change to use the actual task queue length, not the average value

devoncarew:
Change to use the actual task queue length, not the average value; related to https://github.com/dart-lang/sdk/issues/25061.

@pq


pq:
LGTM!

PR #632, opened 2015-12-09
Fix fonts when accessing the site over HTTPS

jsayol:
When accessing the site over HTTPS, the browser blocks the insecure request to http://fonts.googleapis.com


devoncarew:
Thanks!

PR #678, opened 2016-01-08
do not hardcode enableSuperMixins - fixes #677

### LINKED ISSUE #677: superMixin support should not be enabled by default
There is a test to ensure that superMixin support can be enabled and disabled. Both the [enable and disable tests](https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/test/integration/analysis/error_test.dart#L39) pass. This feature should be disabled by default, yet when I paste the sample code from the ["disable superMixin support" test](https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/test/integration/analysis/error_test.dart#L39) in a dart file, I don't see any errors/warnings from analysis server.

```
class Test extends Object with C {
  void foo() {}
}
abstract class B {
  void foo();
}
abstract class C extends B {
  void bar() {
    super.foo();
  }
}
```

Apparently enableSuperMixin is hardcoded true in the Atom codebase.

danrubel:
@devoncarew do not hardcode enableSuperMixins now that analysis server supports enableSuperMixins in _embedded.yaml and .analysis_options

### DIFF
--- a/lib/analysis_server.dart
+++ b/lib/analysis_server.dart
@@ -436,9 +436,8 @@ class _AnalysisServerWrapper extends Server {
     _executables.clear();
     execution.setSubscriptions(['LAUNCH_DATA']);
 
-    // Tracking `enableSuperMixins` here: github.com/dart-lang/sdk/issues/23772.
     analysis.updateOptions(new AnalysisOptions(
-      //enableSuperMixins: true
+      // override Analysis Server defaults here
     ));
 
     server.getVersion().then((v) => _logger.info('version ${v.version}'));
--- a/lib/analysis_server.dart
+++ b/lib/analysis_server.dart
@@ -438,7 +438,7 @@ class _AnalysisServerWrapper extends Server {
 
     // Tracking `enableSuperMixins` here: github.com/dart-lang/sdk/issues/23772.
     analysis.updateOptions(new AnalysisOptions(
-      enableSuperMixins: true
+      //enableSuperMixins: true
     ));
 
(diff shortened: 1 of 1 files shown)


devoncarew:
@sethladd, FYI, we may turn off the enableSuperMixins flag, and rely on users enabling it via config files. Does flutter enable this in their embedder.yaml file?

@pq

sethladd:
Looks like it! https://github.com/flutter/flutter/blob/master/packages/flutter/lib/_embedder.yaml#L3

Try it out. The latest Flutter has this file. Things should "just work"

lib/analysis_server.dart
```diff
@@ -438,7 +438,7 @@ class _AnalysisServerWrapper extends Server {
 
     // Tracking `enableSuperMixins` here: github.com/dart-lang/sdk/issues/23772.
     analysis.updateOptions(new AnalysisOptions(
-      enableSuperMixins: true
+      //enableSuperMixins: true
```
devoncarew: Can you remove this line, and the `enableSuperMixins` comment above? Thanks!

devoncarew:
lgtm w/ tweak

lib/analysis_server.dart
danrubel: Looks like we don't need the entire analysis.updateOptions(...) line, so should I remove that as well? Or keep it with an "override options here" comment?

devoncarew: We can remove it entirely.

danrubel: Done

pq:
What Seth said! :+1:

Issue #703, opened 2016-01-15
re-enable super-mixins support


devoncarew:
This _embedded.yaml file option doesn't work for people on the stable sdk.

PR #704, opened 2016-01-15
re-enable super mixins

devoncarew:
fix https://github.com/dart-atom/dartlang/issues/703

@danrubel, the _embedder.yaml file doesn't work for stable sdks. We can remove this again once the dev sdk becomes the stable one.


danrubel:
Ah! Good point good catch. Sorry I was too quick to remove it.

danrubel:
LGTM

PR #713, opened 2016-01-21
Adds info on how to setup grind.

steelstyle:
Should explicitly say how to call `grind` to people new to dart and `pub`.

Dart sdk installation might also be explicitly specified.


devoncarew:
Thanks!

PR #720, opened 2016-01-25
watch for devices via the flutter daemon

devoncarew:
Watch for devices via the flutter daemon.

- plumb in communication with the `flutter daemon` service
- listen for new devices, device state changes, and device removal

We still need to expose these devices to the user via some UI, and give them the ability to select which device an app will launch on.

@danrubel


lib/flutter/flutter_daemon.dart
```diff
@@ -0,0 +1,373 @@
...
+
+  // TODO: result
+  // TODO: We need the stdout and stderr from launching the process.
+  Future<dynamic> start(
+    String projectDirectory, {
+    String target,
+    bool checked,
+    String route
+  }) {
+    return _call('app.start', _mapify({
```
danrubel: Why create a copy of the map? Is the map modified as part of the call?

lib/flutter/flutter_daemon.dart
```diff
@@ -0,0 +1,373 @@
...
+class RequestError {
+  final String methodName;
+  final dynamic error;
+
+  RequestError(this.methodName, this.error);
+
+  String toString() => '[RequestError ${error}]';
+}
+
+Map _mapify(Map m) {
```
danrubel: Would it make sense to replace calls to this method with calls to Map.from(...) or does this method do something different that's important to its callers?

lib/flutter/flutter_daemon.dart
```diff
@@ -0,0 +1,373 @@
...
+  FlutterSdkManager get _sdkManager => deps[FlutterSdkManager];
+
+  void _initSdk(FlutterSdk sdk) {
+    if (!_enableNotifications) return;
+
+    if (sdk == null) {
+      if (_daemon != null) {
+        _logger.info('Stopping Flutter daemon server');
+      }
+      _daemon?.dispose();
```
danrubel: Move this and the following line into the if (_daemon != null) block?

danrubel:
LGTM

lib/flutter/flutter_daemon.dart
devoncarew: I can rename this method, but what `_mapify` does is return a copy of the map with all keys stripped out that had associated `null` values. So a call site can create a map w/ all possible keys, and `_mapify` will strip out any keys that didn't have associated values.

lib/flutter/flutter_daemon.dart
devoncarew: Different :) I'll rename to clarify it's behavior.

lib/flutter/flutter_daemon.dart
devoncarew: done

lib/flutter/flutter_daemon.dart
devoncarew: renamed to `_stripNullValues`

Issue #701, opened 2016-01-14
send in real file deltas to the analysis server


devoncarew:
https://github.com/dart-atom/dartlang/blob/master/lib/utils.dart#L293

PR #729, opened 2016-01-26
optimize the diff algorithm

devoncarew:
Fix https://github.com/dart-atom/dartlang/issues/701, send in minimal diffs for file edits. Instead of sending in entire file contents for each overlay call, just send in the minimal diff (char additions, char removals).

@danrubel


danrubel:
LGTM

PR #737, opened 2016-01-29
Use $SHELL when invoking shell commands

### LINKED ISSUE #467: support running on the mac with shells besides just bash
like zsh :)

ianloic:
Fixes #467, I think.


lib/flutter/flutter_sdk.dart
```diff
@@ -232,23 +232,12 @@ class FlutterTool {
...
-      return runner;
-    } else {
-      ProcessRunner runner = new ProcessRunner(toolPath, args: args, cwd: cwd);
-      if (startProcess) runner.execStreaming();
-      return runner;
-    }
+  ProcessRunner runRaw(List<String> args,
+      {String cwd, bool startProcess: true}) {
+    ProcessRunner runner =
+        new ProcessRunner.underShell(toolPath, args: args, cwd: cwd);
```
devoncarew: We should still just run `ProcessRunner.underShell()` here on the mac. We don't have the shell issues on linux, and it's likely that `underShell()` will not work on windows (or always emit the warning). You could also move the system check into `underShell()` itself.

lib/flutter/flutter_sdk.dart
```diff
@@ -280,21 +269,10 @@ class _FlutterToolJob extends Job {
...
-          '/bin/bash', args: ['-l', '-c', arg], cwd: cwd);
-      runner.execStreaming();
-      return runner;
-    } else {
-      ProcessRunner runner = new ProcessRunner(
-          sdk.flutterToolPath, args: _args, cwd: cwd);
-      runner.execStreaming();
-      return runner;
-    }
+    ProcessRunner runner = new ProcessRunner.underShell(sdk.flutterToolPath,
```
devoncarew: ditto here re: keeping the check or moving it into `underShell`.

lib/process.dart
```diff
@@ -38,6 +39,33 @@ class ProcessRunner {
...
+    final String shell = utils.env('SHELL');
+
+    if (shell == null) {
+      _logger.warning("Couldn't identify the user's shell");
+      return new ProcessRunner(command, args: args, cwd: cwd, env: env);
+    }
+
+    if (args != null) {
+      // Escape the arguments:
+      List<String> escaped = args.map((String arg) {
```
devoncarew: You should declare this as `Iterable<String>` to avoid problems if this ever ran in checked mode :)

lib/sdk.dart
```diff
@@ -230,20 +230,10 @@ class Sdk {
...
-      ProcessRunner runner = new ProcessRunner(
-          '/bin/bash', args: ['-l', '-c', arg], cwd: cwd);
-      if (startProcess) runner.execStreaming();
-      return runner;
-    } else {
-      ProcessRunner runner = new ProcessRunner(command, args: args, cwd: cwd);
-      if (startProcess) runner.execStreaming();
-      return runner;
-    }
+    ProcessRunner runner =
```
devoncarew: ditto re: the platform check

devoncarew:
Thanks much for taking this on! It was a problem users had encountered that I didn't have a good solution for. I think it makes sense to keep the SHELL check mac specific; let me know what you think.

devoncarew:
(Note to self: we'll want to apply this fix to https://github.com/dart-atom/atom.dart/blob/master/lib/node/process.dart as well)

devoncarew:
Landed - thanks again!

PR #748, opened 2016-02-05
Permits to filter errors on different scope level

steelstyle:
**Context**
On middle size to large size projects, the errors views can have too much information of directory/files the developper is not currently working on.

**PR approach**
This PR gives the ability to switch scope: 
 - Errors in the whole project(default)
 - Errors in the file selected in the editor.
 - Errors in a custom directory.

Possible enhancement:
 - Use a regex instead of a custom directory path, that would allow power user to do more customization.
 - Also filter errors in the `DartLinterConsumer` so we can receive data hidden because of the cap. (Or remove the cap and put it in the `ErrorController`)
- Save this options by project.

This PR is mainly a first attempt to have feedback.

### DIFF
--- a/lib/impl/errors.dart
+++ b/lib/impl/errors.dart
@@ -12,6 +12,7 @@ import '../atom_statusbar.dart';
 import '../atom_utils.dart';
 import '../elements.dart';
 import '../linter.dart';
+import '../projects.dart';
 import '../state.dart';
 import '../utils.dart';
 import '../views.dart';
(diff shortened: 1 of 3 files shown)


devoncarew:
Thanks for the PR! Will take a look shortly.

steelstyle:
Thanks, just for information this is mainly to have a material to discuss on. I would like to land the feature with all requirements once we can agree on the needs.

steelstyle:
@devoncarew any update on this ?

PR #756, opened 2016-02-10
Device toolbar

devoncarew:
Add a toolbar to atom to show the currently connected devices and the applications (for the current project) that are available to run.

screen shot 2016-02-10 at 12 58 52 pm

screen shot 2016-02-10 at 12 58 56 pm

- the device selector only appears if the flutter sdk has been configured
- introduce a `ProjectLaunchManager` class to track the currently active editor, and update and notify when the list of launchable things change
- introduce a `RunnableConfig` abstraction, which is either something which is a potential launch (`Launchable` - a tuple of a runnable file and launch type), or a `LaunchConfiguration` (something which the user has already run at least once)
- allow the user to select apps to run, run an app, and open the configuration for one, from the toolbar
- when a flutter app runs, it'll use the selected device (if any)

@danrubel


lib/flutter/flutter_daemon.dart
```diff
@@ -51,13 +63,11 @@ class FlutterDaemonManager implements Disposable {
         .where((String str) => str.startsWith('[') && str.endsWith(']'))
         .map((String str) => str.substring(1, str.length - 1));
 
-      // process.onStderr.listen((String str) {
-      //   _logger.fine('stderr: $str');
-      // });
-
-      // TODO: Do we want to dispose on exit? Notify someone?
       process.onExit.then((_) {
         _daemon.dispose();
```
danrubel: It appears that _initSdk above calls ```_daemon.dispose()``` if ```_daemon != null``` and then reinitializes _daemon. If that happens, will this code be disposing the new instance of _daemon ?

danrubel:
LGTM

lib/flutter/flutter_daemon.dart
devoncarew: It looks like there was an issue if we received two non-null sdk events in a row. We'd initialize the daemon the first time, and then initialize a 2nd one for the 2nd event, leaving the first one orphaned. I've added a guard check to make sure we only spin up a daemon if there is no current one.

PR #758, opened 2016-02-11
use --start-paused

devoncarew:
Use the `--start-paused` flag when starting Flutter apps so users can hit breakpoints early in the app life-cycle.

@danrubel


danrubel:
LGTM

PR #771, opened 2016-02-17
use /bin/cache/dart-sdk to help locate the dart sdk

devoncarew:
- when trying to auto-locate the Dart SDK, use $FLUTTER_ROOT/bin/cache/dart-sdk as one of the places to look
- fix https://github.com/flutter/atom-flutter/issues/50

@danrubel


lib/sdk.dart
```diff
@@ -292,6 +290,15 @@ class SdkDiscovery {
...
+  /// either exist.
+  String _tryFlutterSdk() {
+    if (!_flutterSdkManager.hasSdk) return null;
+    return _flutterSdkManager.sdk.dartSdkPath;
+  }
+
+  FlutterSdkManager get _flutterSdkManager => deps[FlutterSdkManager];
+
   String _resolveSdkFromVm(String vmPath) {
     if (vmPath == null) return null;
```
danrubel: Does ```which``` return ```null``` or empty string if the path cannot be found?
Do we need to check for empty string as well as ```null```?

danrubel:
LGTM

lib/sdk.dart
devoncarew: `which` will throw if it can't locate the executable - it'll always return a non-null String. The null check is a bit defensive, but it makes sense to add an empty string check; will do.

eseidelGoogle:
So this will mean we can update our instructions so that users will only need to set FLUTTER_ROOT in atom and not bother with DART_SDK?

devoncarew:
Yes, but I'd like to let this sit for a few days. Definitely a goal for the next hackathon.

iposva-google:
I occasionally would like to use a different Dart SDK for non-Flutter based Dart development. Auto-detection is fine if no other is provided, but we should allow this to be overridden.

devoncarew:
The current algorithm is something like:

- if there's an explicitly set dart-sdk, use that
- search the user's path for dart (`which dart`), use the first valid sdk we find
- if there's an explicitly set flutter-root, and it contains a dart sdk, use that
- search the user's path for flutter (`which flutter`); if we find one, and it contains a dart sdk, use that

So we prefer to use explicit settings, and prefer the user's set up PATH, over using the dart sdk in flutter root.

devoncarew:
You can explicitly set the sdk path from the 'dartlang' settings page.

PR #772, opened 2016-02-17
expose flutter:doctor

devoncarew:
@danrubel, companion to this PR: https://github.com/flutter/atom-flutter/pull/52. Expose `flutter doctor` through Atom.


lib/flutter/flutter_tools.dart
```diff
@@ -78,19 +83,35 @@ class FlutterToolsManager implements Disposable {
...
+    }
+
+    FlutterTool flutter = _flutterSdk.sdk.flutterTool;
+
+    TextEditor editor = atom.workspace.getActiveTextEditor();
+    DartProject project = projectManager.getProjectFor(editor?.getPath());
+
+    flutter.runInJob(['doctor'],
+      title: 'Running Flutter doctor…',
+      cwd: project?.directory?.path
```
danrubel: OK to run this outside a flutter project? Good by me if that makes sense. Otherwise might want to not run flutter doctor, disable action, show notification, or something like that.

danrubel:
LGTM

lib/flutter/flutter_tools.dart
devoncarew: Yeah, right now it's not expecting to be in a project dir. It's just testing to see what's installed on the platform (xcode, adb, ...).

I add this here more defensively in case `doctor` does start checking the local project dir.

Perhaps I should instead always have the cwd be the flutter root? Then it'd always be running from the same dir, instead of whatever the cwd of atom happened to be at the time.

danrubel: Another possibility is to change that ```cwd``` line to something like this so that you know where it will run but also support future enhancements to doctor checks of the current project.
```
cwd: project?.directory?.path ?? flutterRootDir
```

devoncarew: I think I'll change it so that the cwd is always the flutter root, and can refactor if the assumptions for `flutter doctor` change.

devoncarew:
merging, thanks for the review!

PR #773, opened 2016-02-17
Add support for async stepping

johnmccutchan:
- Require ^0.2.1 of vm_service_libs
- Update isolate's suspended state based on the pause event's atAsyncSuspension field
- Add stepOverAsyncSuspension
- Add autoStepOver
- Wired UI stepOver to use autoStepOver


johnmccutchan:
This is backwards compatible in that older versions of the VM won't send the "atAsyncSuspension" bit and then the client won't attempt to do a step 'OverAsyncSuspension'

pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: atom_dartlang
-version: 0.6.0
+version: 0.6.1
```
devoncarew: Can you drop this back down? We have a more complicated dance for rev'ing an atom plugin.

johnmccutchan: Done.

lib/debug/observatory_debugger.dart
```diff
@@ -496,7 +505,15 @@ class ObservatoryIsolate extends DebugIsolate {
   stepIn() => service.resume(isolateRef.id, step: StepOption.kInto);
   stepOver() => service.resume(isolateRef.id, step: StepOption.kOver);
   stepOut() => service.resume(isolateRef.id, step: StepOption.kOut);
-
+  stepOverAsyncSuspension() =>
+      service.resume(isolateRef.id, step: StepOption.kOverAsyncSuspension);
```
devoncarew: Ok, so this will only send `kOverAsyncSuspension` into a newer client, as older clients will cause the `atAsyncSuspension` field to populate w/ false above.

lib/debug/observatory_debugger.dart
```diff
@@ -473,12 +478,16 @@ class ObservatoryIsolate extends DebugIsolate {
       (libraryRef) => new ObservatoryLibrary._(libraryRef)).toList();
   }
 
-  void _suspend(bool value) {
-    if (!value) frames = null;
+  void _suspend(bool paused, bool pausedAtAsyncSuspension) {
```
devoncarew: Perhaps used a named param for the 2nd param, just to avoid having call sites w/ a list of booleans (`_suspend(false, false)`) - can be hard to parse later.

devoncarew:
lgtm, thanks! One suggestion about using a named param for clarity.

lib/debug/observatory_debugger.dart
johnmccutchan: Done.

PR #779, opened 2016-02-19
enable and disable the outline button based on the current selection

devoncarew:
- add tooltips to the debugger control buttons
- enable and disable the outline view button in the toolbar based on the current active editor
- remove unused code in the toolbar

@danrubel

### DIFF
--- a/lib/impl/toolbar_impl.dart
+++ b/lib/impl/toolbar_impl.dart
@@ -96,8 +96,8 @@ class DartToolbarContribution implements Disposable {
       runnables = projectLaunchManager.runnables;
 
       runButton.enabled = runnables.isNotEmpty;
-      selectList.disabled = runnables.isEmpty;
-      configureButton.disabled = runnables.isEmpty;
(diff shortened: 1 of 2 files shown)


lib/impl/toolbar_impl.dart
```diff
@@ -114,14 +95,15 @@ class DartToolbarContribution implements Disposable {
     var updateUI = ([_]) {
       runnables = projectLaunchManager.runnables;
 
-      element.disabled = runnables.isEmpty;
-      configureButton.disabled = runnables.isEmpty;
+      runButton.enabled = runnables.isNotEmpty;
       selectList.disabled = runnables.isEmpty;
```
danrubel: nit: it would be easier to parse if all the assignments were ```.enabled``` on the LHS

danrubel:
LGTM

lib/impl/toolbar_impl.dart
devoncarew: Will switch

PR #782, opened 2016-02-19
analyze dartino projects

danrubel:
* Detect Dartino projects by the presence of ```dartino.yaml```
* For any Dartino project that does not contain a ```.packages``` file, pass the default package spec file to the analyzer via setAnalysisRoots

@devoncarew


devoncarew:
lgtm

Issue #705, opened 2016-01-15
Feedback from UX review


sethladd:
- [ ] gray text on blue background is hard to read
- [ ] re-evaluate displaying the detailed architecture in the debugger
- [ ] re-evaluate what we are trying to convey with "Isolate test.dart$main is paused"
- [ ] make "which isolate is selected and running" close to/the same as "how to change isolate"
- [ ] consider using a minus instead of x for "remove/delete breakpoint"

PR #784, opened 2016-02-21
misc debugger UI improvements

devoncarew:
Some work towards https://github.com/dart-atom/dartlang/issues/705 - various improvements to the debugger UI.

- don't show toasts when breakpoints are added or removed (@lukechurch)
- show the device name that we're connecting to in the debugger UI (instead of the processor and OS information)
- make isolate names slightly more user friendly
- use a `-` instead of an `x` to delete a breakpoint
- improve the contrast for selected debugger items
- remove the console clear button - its functionality was misleading

@danrubel 

screen shot 2016-02-20 at 8 35 34 pm

screen shot 2016-02-20 at 8 38 23 pm


lib/debug/observatory_debugger.dart
```diff
@@ -472,6 +472,9 @@ class ObservatoryIsolate extends DebugIsolate {
 
   String get name => isolateRef.name;
 
+  String get displayName =>
+    name.contains(r'$') ? "${name.replaceAll(r'$', ':')}()" : name;
```
danrubel: Could you add a comment as to why the `replaceAll` is needed?
And why append `()` to the displayName if it contains `$`?

danrubel:
LGTM

lib/debug/observatory_debugger.dart
devoncarew: Will do. Isolate names are something like `foo_app.dart$main`. This cleans them up slightly to `foo_app.dart:main()`; I'll add a note.

PR #787, opened 2016-02-23
various debugger improvements

devoncarew:
Various debugger improvements:

- make sure an isolate is selected, even if there's not one currently paused
- tweak the resume button so that it doesn't look like the launch button
- show the isolate status next to the isolate name (_running_, _paused (no frames)_); for paused, don't show the state
- have the resume button switch between resume and pause modes
- add a right-click context menu item to toggle breakpoints
- give more space to the list of stack frames
- de-bounce the UI when stepping; this reduces the amount of flashing in the UI and makes the UI appear more solid

@danrubel


danrubel:
LGTM

PR #788, opened 2016-02-23
fix visibility of the devices pulldown on the toolbar

devoncarew:
Fixed an issue with the devices visibility on the toolbar.

@danrubel


danrubel:
LGTM

PR #789, opened 2016-02-23
refactor dartlang to use atom.dart fs

danrubel:
This removes File, Directory, Entry, and the like from dartlang in favor of identical functionality in atom.dart.

@devoncarew

### DIFF
--- a/lib/atom.dart
+++ b/lib/atom.dart
@@ -3,7 +3,6 @@
 // BSD-style license that can be found in the LICENSE file.
 
 /// A Dart wrapper around the [Atom](https://atom.io/docs/api) apis.
-library atom2;
 
 import 'dart:async';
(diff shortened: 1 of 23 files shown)


lib/atom.dart
```diff
@@ -3,12 +3,13 @@
 // BSD-style license that can be found in the LICENSE file.
 
 /// A Dart wrapper around the [Atom](https://atom.io/docs/api) apis.
-library atom;
+library atom2;
```
devoncarew: lgtm; could also drop the library directive altogether

devoncarew:
lgtm!

lib/atom.dart
danrubel: Good point.

devoncarew:
(landing to resolve some merge conflicts)

PR #800, opened 2016-02-25
de-emphasized Mojo launches

devoncarew:
De-emphasize Mojo launches; most users will be running Flutter apps (fix https://github.com/flutter/flutter/issues/2006). We'll show existing mojo launches, but not offer to create new ones.

Eventually we'll re-write the launch configs a bit so we show all existing launch configs, and give the user a more explicit way to create new ones.

@danrubel


devoncarew:
@danrubel, landing TBR to we can release

danrubel:
LGTM

PR #807, opened 2016-02-26
Point devs to how to set args when launching

sethladd:


sethladd:
Not sure if the formatting will work for comments. But here it is :)

cc @devoncarew

lib/flutter/flutter_launch.dart
```diff
@@ -70,7 +70,7 @@ class FlutterLaunchType extends LaunchType {
   }
 
   String getDefaultConfigText() {
-    return 'checked: true\ndebug: true\nroute:\n';
+    return 'checked: true\ndebug: true\nroute:\n# additional args for the flutter run command\nargs:\n';
```
devoncarew: ==> `# Additional args for the flutter run command:`

Also, we may want to break this into a multi-line string for readability (now that it's gotten a bit longer).

devoncarew:
Thanks! One slight tweak to the text.

sethladd:
fixed!

Issue #805, opened 2016-02-26
Uncaught RangeError (index): Index out of range: index must not be negative: -2


jimbeveridge:
[Enter steps to reproduce below:]

(I have no idea what I did to cause this.)

**Atom Version**: 1.5.3
**System**: Ubuntu 14.04.1
**Thrown From**: [dartlang](https://github.com/dart-atom/dartlang) package, v0.6.2

### Stack Trace

Uncaught RangeError (index): Index out of range: index must not be negative: -2

```
At /usr/local/google/home/jimbe/.atom/packages/dartlang/web/entry.dart.js:23569

RangeError (index): Index out of range: index must not be negative: -2
    at [object Object].dart.wrapException (/usr/local/google/home/jimbe/.atom/packages/dartlang/web/entry.dart.js:24633:17)
    at [object Object].dart.ioore (/usr/local/google/home/jimbe/.atom/packages/dartlang/web/entry.dart.js:24586:15)
    at _MatchParser.dart._MatchParser.parseMatch$1 (/usr/local/google/home/jimbe/.atom/packages/dartlang/web/entry.dart.js:40464:20)
    at FindReferencesView.dart.FindReferencesView._references$_render$2 (/usr/local/google/home/jimbe/.atom/packages/dartlang/web/entry.dart.js:40333:36)
    at [object Object].<anonymous> (/usr/local/google/home/jimbe/.atom/packages/dartlang/web/entry.dart.js:24887:32)
    at ListTreeBuilder.dart.ListTreeBuilder.render$2 (/usr/local/google/home/jimbe/.atom/packages/dartlang/web/entry.dart.js:42615:28)
    at ListTreeBuilder.dart.ListTreeBuilder._addNode$2 (/usr/local/google/home/jimbe/.atom/packages/dartlang/web/entry.dart.js:42559:16)
    at ListTreeBuilder.dart.ListTreeBuilder._addNode$2 (/usr/local/google/home/jimbe/.atom/packages/dartlang/web/entry.dart.js:42581:18)
    at FindReferencesView__handleSearchResults_closure.dart.FindReferencesView__handleSearchResults_closure.call$1 (/usr/local/google/home/jimbe/.atom/packages/dartlang/web/entry.dart.js:40434:14)
    at _RootZone.dart._RootZone.runUnary$2 (/usr/local/google/home/jimbe/.atom/packages/dartlang/web/entry.dart.js:46503:20)
[...]

PR #809, opened 2016-02-26
fix an array out of bounds exception

devoncarew:
Fix https://github.com/dart-atom/dartlang/issues/805. The line pos from the find references analysis server command can be `-1` in some cases. We need to guard against this before trying to index the line array.

@danrubel


danrubel:
LGTM

PR #810, opened 2016-02-27
refactor/style(atom.autocomplete): consistent use of types, ...

thso:
more consistent use of `var` for local variables where inference does the job + extract inner anonymous functions adding return types (it's also faster) + use of `async/await` and nullaware operators where it makes the code cleaner + dartfmt on `dart_autocomplete_provider`.

Also fixed a few type issues discovered when adding the types and enabling strong-mode.


devoncarew:
> extract inner anonymous functions adding return types

Awesome, that completion function had (organically) grown pretty large.

> Also fixed a few type issues discovered when adding the types and enabling strong-mode.

Great! Thanks for the fixes.

FYI, we're annotating types on the left, even inside methods. There are a few outliers in the source base where that's not the case, but we generally avoid `var`.

Also, we're not using the formatter, as we're hewing closer to the Flutter code style. Parenthesis and array braces are balanced (like curly braces) and we're flexible on the 80-col limit when it improves readability.

With the formatting changes and refactorings, it's a bit hard to review this change. Could you re-do this as two PRs, one with the additional types and strong mode fixes, and one with the extracted anonymous methods? That's a pain I know, but given the complexity of the completion code, and how important the completion UI is to a good experience, I want to make sure that we're not regressing something by accident here. Thanks!

thso:
I knew something fishy was going on, as none of the code was `dartfmt`ed :p Splitting is not a problem, thanks for the quick feedback!

PR #812, opened 2016-02-28
refactor(autocomplete): Extract inner anonymous methods + add return types, ...

thso:
`+` minor style changes/simplifications e.g. using null-aware operators.


devoncarew:
lgtm, thanks!

PR #813, opened 2016-02-29
add a type annotation fix

devoncarew:
Add back in a type annotation fix (`List<Suggestion>` ==> `List<JsObject>`) from https://github.com/dart-atom/dartlang/pull/810.

@thso


thso:
Ah, lgtm, but this is also part of #814

devoncarew:
Ah, cool will close this one.

PR #814, opened 2016-02-29
refactor(autocomplete): Add types + async/await, null-aware operators simplifications.

thso:
`+` fix some type issues discovered when adding the types and enabling strong-mode.


devoncarew:
lgtm!

devoncarew:
Also, two things:

- you might want to bump your sdk up to a `1.14` one - you're re-compiling with `1.13.2`
- you don't need to check in the compiled artifacts. I'll generally re-compile them on my machine before a release

thso:
OK, will do, thanks!

PR #815, opened 2016-02-29
add Dartino launch type to Atom

danrubel:
@devoncarew

### DIFF
--- a/lib/dartino/launch_dartino.dart
+++ b/lib/dartino/launch_dartino.dart
@@ -1,0 +1,121 @@
+import 'dart:async';
+
+import 'package:atom/node/fs.dart';
+import 'package:atom/node/process.dart';
+import 'package:logging/logging.dart';
+
+import '../atom.dart';
+import '../launch/launch.dart';
+import '../projects.dart';
+import '../state.dart';
+
+final Logger _logger = new Logger('atom.dartino_launch');
+
+class DartinoLaunchType extends LaunchType {
+  static void register(LaunchManager manager) =>
+      manager.registerLaunchType(new DartinoLaunchType());
+
+  DartinoLaunch _lastLaunch;
+
+  DartinoLaunchType() : super('dartino');
+
+  bool canLaunch(String path) {
+    DartProject project = projectManager.getProjectFor(path);
+    if (project == null || !project.isDartinoProject()) return false;
+    if (!analysisServer.isExecutable(path)) return false;
+    return true;
+  }
+
+  List<String> getLaunchablesFor(DartProject project) {
+    if (project == null || !project.isDartinoProject()) return [];
+    var files = <String>[];
+    visit(Directory dir) {
+      for (Entry entry in dir.getEntriesSync()) {
+        if (entry.isDirectory()) {
+          visit(entry);
+          continue;
+        }
+        var path = entry.getPath();
+        if (path != null && path.endsWith('dart')) files.add(path);
+      }
+    }
+    visit(project.directory);
+    return files;
+    //TODO(danrubel) add Dartino launch support to analysis server
+    // return analysisServer
+    //     .getExecutablesFor(project.path)
+    //     .where((String path) => path.endsWith('dart'))
+    //     .toList();
+  }
+
+  Future<Launch> performLaunch(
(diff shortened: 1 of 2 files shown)


lib/dartino/launch_dartino.dart
```diff
@@ -0,0 +1,121 @@
...
+  DartinoLaunchType() : super('dartino');
+
+  bool canLaunch(String path) {
+    DartProject project = projectManager.getProjectFor(path);
+    if (project == null || !project.isDartinoProject()) return false;
+    if (!analysisServer.isExecutable(path)) return false;
+    return true;
+  }
+
+  List<String> getLaunchablesFor(DartProject project) {
```
devoncarew: This will return all dart files in a dartino project as launch entry points. You may want to instead use the `analysisServer.getExecutablesFor(String path)` call in the analysis server. The `launch_cli.dart` and `flutter_launch.dart` files do this.

lib/dartino/launch_dartino.dart
```diff
@@ -0,0 +1,121 @@
...
+    //     .toList();
+  }
+
+  Future<Launch> performLaunch(
+      LaunchManager manager, LaunchConfiguration configuration) async {
+    String path = configuration.primaryResource;
+
+    DartProject project = projectManager.getProjectFor(path);
+    if (project == null || !project.isDartinoProject()) {
+      throw "File not in a Dartino project.";
```
devoncarew: You may want to verify that the launch framework handles exceptions well. If not, we should change it to handle them.

lib/dartino/launch_dartino.dart
```diff
@@ -0,0 +1,121 @@
...
+    _lastLaunch = new DartinoLaunch(manager, this, configuration);
+    manager.addLaunch(_lastLaunch);
+    // TODO (danrubel) to be implemented
+    // sdk.launch(_lastLaunch);
+    atom.notifications.addError('Not implemented yet');
+    return _lastLaunch;
+  }
+
+  String getDefaultConfigText() {
+    return '';
```
devoncarew: You can put default option values here that the user can edit. Things like checked mode, debug, cli args, ...

lib/dartino/launch_dartino.dart
devoncarew: Nevermind, I think this is handled.

lib/dartino/launch_dartino.dart
danrubel: Yes it will ... see note 10 lines below. This is temporary until I can add support to analysis server. Returning all the files is not an issue at this point because Dartino projects are small. Will update this method once analysis server has the appropriate support for Dartino.

lib/dartino/launch_dartino.dart
devoncarew: Perhaps a TODO: if you want to circle back to this.

lib/dartino/launch_dartino.dart
devoncarew: OK.

There was a change that went into the analysis server a while ago that may be what you want. We can query to get the transitive imports of a library. It'll let the client decide what an application type a library is (web, flutter, dartino, ...) w/o having to add specific support for each type into the analysis server.

devoncarew:
lgtm

lib/dartino/launch_dartino.dart
danrubel: Hmmm... good point. I am following the pattern in [flutter launch](https://github.com/dart-atom/dartlang/blob/master/lib/flutter/flutter_launch.dart#L58) where it throws an exception in the same situation. If that one is handled properly, then I assume this one will be as well. I'll verify and fix as needed in a subsequent PR.

lib/dartino/launch_dartino.dart
danrubel: Good point. Will add in subsequent PR.

lib/dartino/launch_dartino.dart
devoncarew: I dug into the code; the `async` here threw me off. It will be returning an async exception, which we'll handle :)

danrubel:
Merged

PR #816, opened 2016-02-29
add a type annotation; recompile

devoncarew:
- add a type annotation in `dart_autocomplete_provider.dart`
- recompile to capture the latest PRs

cc @thso, @danrubel


danrubel:
LGTM

PR #817, opened 2016-02-29
feat(type-tooltip): Add a tooltip with type information on mouse hover.

thso:
The tooltip appears when mousing over something in an editor and contains the same information as the title bar of `F1 dartdoc` information. It is currently hidden behind an experimental flag `hoverTooltip` set to `false` by default.

The tooltip in action:
tooltip_example

While roughly works, the functionality is here to be discussed and there is some work to be done on the content of the tooltip, the colors (hardcoded to blue and white right now), debouncing, and it currently shows even when the content is empty.

Linked to #7.

### DIFF
--- a/lib/plugin.dart
+++ b/lib/plugin.dart
@@ -68,6 +68,7 @@ import 'usage.dart' show UsageManager;
 import 'utils.dart';
 import 'views.dart';
 import 'views.dart' show ViewGroupManager;
+import 'package:atom_dartlang/impl/tooltip.dart';
 
 export 'atom.dart' show registerPackage;
 
@@ -154,6 +155,7 @@ class AtomDartPackage extends AtomPackage {
     disposables.add(new NavigationHelper());
     disposables.add(new OrganizeFileManager());
     disposables.add(new OutlineController());
+    disposables.add(new TooltipController());
     disposables.add(pubManager);
     disposables.add(runAppManager);
     disposables.add(new RefactoringHelper());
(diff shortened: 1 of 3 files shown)


devoncarew:
Thanks for the PR! I'll take a look at this a little later today.

styles/tooltip.less
```diff
@@ -0,0 +1,14 @@
+#hover-tooltip {
+  background-color: #226AEF;
```
devoncarew: We probably want to key off the current theme's style variables. There are a few examples in our less files.

Perhaps add a TODO: here, or track this in a bug to revise this in a follow-on PR.

lib/plugin.dart
```diff
@@ -388,6 +390,14 @@ class AtomDartPackage extends AtomPackage {
         'type': 'boolean',
         'default': false,
         'order': 11
+      },
+
+      // experimental features
+      'hoverTooltip': {
```
devoncarew: We really only need to add config items here if we want them to be exposed to the user. We can still read (`atom.config.getValue()`) arbitrary keys w/o them being listed here.

devoncarew: We can keep this here though - just FYI.

devoncarew:
This is great, thanks. We should track improvements in an issue. Before we remove the experimental flag, we'll want to:

- debounce, so that it shows when the mouse cursor is quiet for a short period
- borrow style (and content) from the current F1 dialog. It doesn't need to look the same; I think there's room for improvement in that dialog
- ensure that users are able to scroll within a tooltip w/ a lot of content w/o the tooltip disappearing when they move their mouse
- remove the F1 dialog
- remove the experimental flag :)

devoncarew:
lgtm, landing. Can you open an issue w/ the above?

thso:
Done, #825

PR #818, opened 2016-02-29
Autoconfig sdk

devoncarew:
- if we auto-locate a flutter sdk, and there is no dart sdk configured, then use the dart sdk inside the flutter sdk
- remove some older plugin settings

@danrubel


danrubel:
LGTM

PR #820, opened 2016-03-01
utility to download and install sdk

danrubel:
@devoncarew


lib/plugin.dart
```diff
@@ -191,6 +192,7 @@ class AtomDartPackage extends AtomPackage {
       });
     });
     _addCmd('atom-workspace', 'dartlang:send-feedback', (_) => _handleSendFeedback());
+    _addCmd('atom-workspace', 'dartino:installSdk', dartino.promptInstallSdk);
```
devoncarew: 'dartino:install-sdk'?

devoncarew:
Lgtm

I couldn't see - does this clean up the temp dir and SDK zip?

lib/plugin.dart
danrubel: Good idea. Done.

danrubel:
It should but does not. Fixed.

danrubel:
Merged

PR #823, opened 2016-03-01
updated code completion to reduce inadvertent completions after array braces

devoncarew:
Updated code completion to reduce inadvertent completions after array braces.

@thso

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # dartlang plugin changelog
 
+## unreleased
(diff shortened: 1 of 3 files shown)


thso:
lgtm

thso:
Nevermind, we want to keep suggesting after `)` for things like `async`, etc.

devoncarew:
It might make sense to dis-allow `)`; we could still allow `) `. I'll experiment to see how well that works.

PR #826, opened 2016-03-01
feat(autocomplete): Enable fuzzy search of autocomplete suggestions.

thso:
This adds a feature by removing code, how about that?!

Here is what it would look like:
fuzzy_autocomplete


devoncarew:
lgtm!

fix https://github.com/dart-atom/dartlang/issues/708

devoncarew:
@thso, we should probably add an entry in the changelog for this.

Issue #350, opened 2015-09-23
Plugin should auto insert closing for /*


lukechurch:
We auto-close most of the block symbols. E.g. { adds a }, ( adds a ).

Consequently I keep expecting the editor to auto close /* with */ on the next line.

devoncarew:
mifxed

PR #828, opened 2016-03-01
feat(comment-edit): Auto-close comment block when hitting enter after /*.

### LINKED ISSUE #827: Comments starting with '/*' are not autoclosed.
When pressing `Enter` after `/*`, I expect the following to be inserted:
```
/*
 * [Cursor here]
 */
```
instead of the current '*' that's inserted and that leave the rest of the code commented out.

Additionally, there seems to be a bug when the last line ends with `/` and the cursor is moved before the end of the line before pressing `Enter`:
bug_comment_1

thso:
Automatically close the comment block by inserting `*`, `*/` when pressing enter after opening a block comment `/*`.

The new behaviour looks like this:
autoclose_comment

Also added `move_` methods to `TextEditor` façade and removed now obsolete tricks to close the comment by adding a `/` on the last line and hitting enter.

Closes #827

### DIFF
--- a/lib/impl/editing.dart
+++ b/lib/impl/editing.dart
@@ -70,10 +70,8 @@ bool _handleEnterKey(TextEditor editor, int row, int col) {
   if (!inComment) return false;
   if (leading.isEmpty) leading = ' ';
 
-  String previousLine = '';
-  if (row > 0) {
-    previousLine = editor.lineTextForBufferRow(row - 1).trimLeft();
-  }
+  String previousLine = editor.lineTextForBufferRow(row - 1)?.trimLeft() ?? '';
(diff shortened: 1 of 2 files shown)


lib/impl/editing.dart
```diff
@@ -94,10 +92,22 @@ bool _handleEnterKey(TextEditor editor, int row, int col) {
   }
 
   if (trimmedText.startsWith('/*')) {
-    editor.atomic(() {
-      editor.insertNewline();
-      editor.insertText(' * ');
-    });
+    if(nextLine.startsWith('*')) {
```
devoncarew: nit: space after `if`

devoncarew:
lgtm

fix https://github.com/dart-atom/dartlang/issues/350

PR #837, opened 2016-03-03
feat(comment-editing): Continue a normal comment block when splitting an existing comment in…

thso:
…to multiple lines or when hitting enter after a comma.

Looks like:
better_normal_comments

Current experience is quite frustrating as when splitting a comment in two, the last part will become uncommented.

### DIFF
--- a/lib/impl/editing.dart
+++ b/lib/impl/editing.dart
@@ -10,6 +10,11 @@ import '../atom.dart';
 
 final Logger _logger = new Logger('editing');
 
+// TODO: If we're in a line comment, and the line is longer than the max line
+// length, extend the comment.
+// var prefLineLength = atom.config.get('editor.preferredLineLength',
+//   scope: editor.getRootScopeDescriptor());
+
 // TODO: Tests for this.
 
 /// Handle special behavior for the enter key in Dart files. In particular, this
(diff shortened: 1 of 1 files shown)


lib/impl/editing.dart
```diff
@@ -59,6 +54,8 @@ bool _handleEnterKey(TextEditor editor, int row, int col) {
     String temp = trimmedText.substring(3).trimLeft();
     leading = trimmedText.substring(3, trimmedText.length - temp.length);
     inComment = (line.length - trimmedText.length + 2) <= col;
+  } else if(trimmedText.startsWith('//')) {
```
devoncarew: nit: `if(` ==> `if (`

lib/impl/editing.dart
```diff
@@ -91,6 +88,19 @@ bool _handleEnterKey(TextEditor editor, int row, int col) {
     return true;
   }
 
+  if (trimmedText.startsWith('//')) {
```
devoncarew: `if (trimmedText.startsWith('//') && !trimmedText.startsWith('///') {`? Or do you intend this to also work in dartdoc comments?

lib/impl/editing.dart
```diff
@@ -10,11 +10,6 @@ import '../atom.dart';
 
 final Logger _logger = new Logger('editing');
 
-// TODO: If we're in a line comment, and the line is longer than the max line
-// length, extend the comment.
-// var prefLineLength = atom.config.get('editor.preferredLineLength',
```
devoncarew: Let's keep this TODO: My idea here is that if the comment line (`//`) is longer than 80 cols, and they hit enter, that we should backtrack back to the first ws before 80 cols, wrap the line, and continue the comment. So they can write long lines w/o having to manually know when to wrap a line.

devoncarew:
This is great! Lots of comments, but does lgtm. Would like to address the nits before landing.

lib/impl/editing.dart
thso: If the line started with '///', I believe it would have been handled by the if preceding this one.

lib/impl/editing.dart
thso: Interesting, will work on making that work too.

thso:
@devoncarew PTAL

devoncarew:
lgtm!

PR #839, opened 2016-03-04
integrate Dartino launching into dartlang framework

danrubel:
* integrated Dartino app launch into dartlang launching framework
* use 'dartino flash' command to compile and deploy rather than specialized Atom specific code
* show command line details in console during launch

@devoncarew

### DIFF
--- a/lib/dartino/device/device.dart
+++ b/lib/dartino/device/device.dart
@@ -1,0 +1,47 @@
+import 'dart:async';
+
+import 'package:atom_dartlang/atom.dart';
+
+import '../dartino_util.dart';
+import '../launch_dartino.dart';
+import '../sdk/dartino_sdk.dart';
+import 'stm32f746disco.dart';
+
+/// The connected device on which the application is executed.
+abstract class Device {
+  /// Return a target device for the given launch.
+  /// If there is a problem or a compatible device cannot be found
+  /// then notify the user and return `null`.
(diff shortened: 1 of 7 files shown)


lib/dartino/device/device.dart
```diff
@@ -0,0 +1,47 @@
+import 'dart:async';
+
+import 'package:atom_dartlang/atom.dart';
+
+import '../dartino_util.dart';
+import '../launch_dartino.dart';
+import '../sdk/dartino_sdk.dart';
+import 'stm32f746disco.dart';
```
devoncarew: Perhaps a slightly less exact but more user friendly file name?

devoncarew:
lgtm

lib/dartino/device/device.dart
danrubel: Point taken, but these are the names used in the dartino project, and I want to keep them the same in both atom and in dartino for my own sanity.

devoncarew: Makes sense -

danrubel: I'm hoping that after a few boards, I can generalize, and then yes, the names will be more user friendly at that point :-)

danrubel:
Merged https://github.com/dart-atom/dartlang/commit/7fe56f9524082cd9c309531838535632176a92bd

PR #840, opened 2016-03-04
refactor to use atom.dart disposables

danrubel:
@devoncarew


devoncarew:
That's a big one! lgtm

PR #845, opened 2016-03-05
feat(comment-editing): Automatically wrap a comment that is longer than ...

thso:
...the configured preferred line length + added several `moveTo_` methods to `TextEditor`.

comment_wrapping

As a follow-up to this, after cleaning up the logic, I will add the same wrapping behaviour to the other kind of comments.

### DIFF
--- a/lib/impl/editing.dart
+++ b/lib/impl/editing.dart
@@ -10,11 +10,6 @@ import '../atom.dart';
 
 final Logger _logger = new Logger('editing');
 
-// TODO: If we're in a line comment, and the line is longer than the max line
-// length, extend the comment.
-// var prefLineLength = atom.config.get('editor.preferredLineLength',
-//   scope: editor.getRootScopeDescriptor());
-
 // TODO: Tests for this.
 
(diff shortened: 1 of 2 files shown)


lib/impl/editing.dart
```diff
@@ -101,6 +96,22 @@ bool _handleEnterKey(TextEditor editor, int row, int col) {
       });
 
       return true;
+    } else if (atEol) {
+      var prefLineLength = atom.config
+          .getValue('editor.preferredLineLength', scope: editor.getRootScopeDescriptor());
+
+      if (col >= prefLineLength) {
+        int wrapAtCol = line.substring(0, prefLineLength + 1).lastIndexOf(' ');
```
devoncarew: Can you be a bit defensive here? If `wrapAtCol` == -1, we should not wrap, but still continue the comment.

devoncarew:
lgtm modulo being defensive when looking for a place to wrap. I've been wanting this for a long time! Very happy to see it landing.

thso:
@devoncarew Done! Ptal.

Issue #801, opened 2016-02-25
change our error caping mechanism to only work per-project


devoncarew:
Right now we cap the number of errors per-project, and per-session (all projects combined). We should just cap per-project, so one project with a lot of errors won't affect the other projects.

apwilson:
Some context.  I have ~10 projects open at a time.  If one of those projects has > 500 errors, no errors are ever shown for the remaining ~9 projects which gives the impression the analyzer has stopped running.  One bad project shouldn't ruin the bunch.

PR #849, opened 2016-03-08
adjust the max issue cap

devoncarew:
- make the depth we recurse looking for dart projects a little deeper (from 2 directories to 4)
- adjust the way we calculate the max number of issues. Before this, we had a per-file limit and a per-workspace limit. This would cause projects w/ lots of errors to hide errors on projects w/ only a few. We now have a per-file limit and a per project limit (200, and 500). Fix https://github.com/dart-atom/dartlang/issues/801

@danrubel


danrubel:
LGTM

PR #852, opened 2016-03-09
upgrade the analysis server lib version

devoncarew:
- upgrade to the latest analysis server spec, that captures the new error code field
- recompile the JS output

@danrubel


danrubel:
LGTM

PR #854, opened 2016-03-10
hide EventListener

danrubel:
I'm removing `dartlang/lib/js.dart` in favor of `atom.dart/lib/src/js.dart` and in the process moving EventListener into `atom.dart/lib/utils/disposables.dart`. This precursor CL prevents dartlang from trying to import EventListener from both old and new locations. This `hide EventListener` will be removed as part of removing `js.dart` from dartlang.
@devoncarew


devoncarew:
lgtm!

danrubel:
merged

Issue #833, opened 2016-03-02
Uncaught 105: Extension errorIsolate must be runnable before this request is made.


jimbeveridge:
[Enter steps to reproduce below:]

I just clicked Run. I don't remember doing anything strange.

**Atom Version**: 1.5.3
**System**: Ubuntu 14.04.1
**Thrown From**: [dartlang](https://github.com/dart-atom/dartlang) package, v0.6.4

### Stack Trace

Uncaught 105: Extension error
Isolate must be runnable before this request is made.

```
At /usr/local/google/home/jimbe/.atom/packages/dartlang/web/entry.dart.js:23569

105: Extension error
Isolate must be runnable before this request is made.
    at [object Object].dart.wrapException (/usr/local/google/home/jimbe/.atom/packages/dartlang/web/entry.dart.js:24633:17)
    at _rootHandleUncaughtError_closure.dart._rootHandleUncaughtError_closure.call$0 (/usr/local/google/home/jimbe/.atom/packages/dartlang/web/entry.dart.js:46425:19)
    at [object Object].dart._microtaskLoop (/usr/local/google/home/jimbe/.atom/packages/dartlang/web/entry.dart.js:43837:27)
    at [object Object].dart._startMicrotaskLoop (/usr/local/google/home/jimbe/.atom/packages/dartlang/web/entry.dart.js:43843:11)
    at TimerImpl_internalCallback0.dart.TimerImpl_internalCallback0.call$0 (/usr/local/google/home/jimbe/.atom/packages/dartlang/web/entry.dart.js:23948:23)
    at invokeClosure_closure.dart.invokeClosure_closure.call$0 (/usr/local/google/home/jimbe/.atom/packages/dartlang/web/entry.dart.js:25836:29)
    at _IsolateContext.dart._IsolateContext.eval$1 (/usr/local/google/home/jimbe/.atom/packages/dartlang/web/entry.dart.js:23560:25)
    at [object Object].dart._callInIsolate (/usr/local/google/home/jimbe/.atom/packages/dartlang/web/entry.dart.js:23202:28)
    at dart.invokeClosure (/usr/local/google/home/jimbe/.atom/packages/dartlang/web/entry.dart.js:24781:20)
    at /usr/local/google/home/jimbe/.atom/packages/dartlang/web/entry.dart.js:24802:18
```

### Commands

```
[...]

Issue #808, opened 2016-02-26
Uncaught 105: Extension errorIsolate must be runnable before this request is made.


sethladd:
[Enter steps to reproduce below:]

1. Cmd-R to run a Flutter app. This is the second time I ran my app, after updating to the latest Flutter master branch.

**Atom Version**: 1.5.3
**System**: Mac OS X 10.11.3
**Thrown From**: [dartlang](https://github.com/dart-atom/dartlang) package, v0.6.2

### Stack Trace

Uncaught 105: Extension error
Isolate must be runnable before this request is made.

```
At /Users/sethladd/.atom/packages/dartlang/web/entry.dart.js:23569

105: Extension error
Isolate must be runnable before this request is made.
    at [object Object].dart.wrapException (/Users/sethladd/.atom/packages/dartlang/web/entry.dart.js:24633:17)
    at _rootHandleUncaughtError_closure.dart._rootHandleUncaughtError_closure.call$0 (/Users/sethladd/.atom/packages/dartlang/web/entry.dart.js:46425:19)
    at [object Object].dart._microtaskLoop (/Users/sethladd/.atom/packages/dartlang/web/entry.dart.js:43837:27)
    at [object Object].dart._startMicrotaskLoop (/Users/sethladd/.atom/packages/dartlang/web/entry.dart.js:43843:11)
    at TimerImpl_internalCallback0.dart.TimerImpl_internalCallback0.call$0 (/Users/sethladd/.atom/packages/dartlang/web/entry.dart.js:23948:23)
    at invokeClosure_closure.dart.invokeClosure_closure.call$0 (/Users/sethladd/.atom/packages/dartlang/web/entry.dart.js:25836:29)
    at _IsolateContext.dart._IsolateContext.eval$1 (/Users/sethladd/.atom/packages/dartlang/web/entry.dart.js:23560:25)
    at [object Object].dart._callInIsolate (/Users/sethladd/.atom/packages/dartlang/web/entry.dart.js:23202:28)
    at dart.invokeClosure (/Users/sethladd/.atom/packages/dartlang/web/entry.dart.js:24781:20)
    at /Users/sethladd/.atom/packages/dartlang/web/entry.dart.js:24802:18
```

### Commands

```
  2x -1:04.0 core:backspace (atom-text-editor.editor.is-focused)
     -1:03.6.0 core:save (atom-text-editor.editor.is-focused)
     -1:01.8.0 dartlang:run-application (atom-text-editor.editor.is-focused)
[...]

PR #868, opened 2016-03-14
fix an Isolate must be runnable exception

devoncarew:
- fixed an issue with debugger startup (showing an 'Isolate must be runnable' exception)
- fix https://github.com/dart-atom/dartlang/issues/833, https://github.com/dart-atom/dartlang/issues/808

Under some conditions, we get an `kPauseStart` event (the isolate was paused at application start) before we get a `kIsolateRunnable` (ready to resume the isolate)

@danrubel


danrubel:
LGTM

PR #871, opened 2016-03-16
adjust rebuild plugin functionality - fixes #870

### LINKED ISSUE #870: Open bracket ([) character invokes rebuild of atom plugins
This issue happens on a swiss-german keyboard layout and only if I have dartlang installed in atom.

atom version: 1.5.4
package version (dartlang): 0.6.6 
running on windows 7 pro 64 bit

pressing the keystroke on an empty file the following notifications show up:

atom_dartlang

pressing the keystroke in a file which has content, reacts like this:

atom_dartlang2

By removing the package, the problem is resolved. But that's surely not my intent. Maybe the following information might be helpful:

I have to press "ALT-GR + open bracket" (right ALT) to type that character.

danrubel:
modify the rebuild plugin functionality to run *only* if the `buildAtomPlugins` setting is specified.
@devoncarew 

fix #870

### DIFF
--- a/lib/impl/rebuild.dart
+++ b/lib/impl/rebuild.dart
@@ -19,8 +19,11 @@ class RebuildManager implements Disposable {
   Disposables disposables = new Disposables();
 
   RebuildManager() {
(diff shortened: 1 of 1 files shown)


devoncarew:
lgtm

PR #872, opened 2016-03-16
rewrite the launchable app pulldown and run button code

devoncarew:
@danrubel, re-write our app pulldown and launch selection code; this PRs slightly large and hits many files -

- show all existing launch configurations for the workspace
- also show any launches specific to the current file
- the currently selected launch is sticky - we don't change it as the active editor changes
- simply launch logic to only detect based on the contents of the current file (`main(`), and not use the analysis server to find all launchable resources for the project
- show the devices pulldown if the currently selected launch type supports it
- increase the width of the pulldown; the launch names now include the project name
- make the run button and cmd-R both behave the same - they both will launch what's indicated in the app pulldown

In a follow-on PR, I'll update to temporarily change the pulldown selection if the current file is runnable. Else it'll default to to the last user selection, or the last launched file if there was no selection.


lib/flutter/flutter_launch.dart
```diff
@@ -30,7 +30,9 @@ class FlutterLaunchType extends LaunchType {
 
   String get flutterRunCommand => 'run';
 
-  bool canLaunch(String path) {
+  bool canLaunch(String path, { LaunchData data }) {
+    if (data == null) return false;
```
danrubel: If data is required to launch, then why make the argument optional?

danrubel: Ah! I found it. Other implementations don't need it. All good.

lib/flutter/flutter_launch.dart
```diff
@@ -38,18 +40,7 @@ class FlutterLaunchType extends LaunchType {
     if (!analysisServer.isExecutable(path)) return false;
```
danrubel: Should this line be removed?

lib/launch/launch_configs.dart
```diff
@@ -38,6 +42,12 @@ class LaunchConfigurationManager implements Disposable, StateStorable {
     return _getCreateProjectConfig(path).getConfigs();
   }
 
+  List<LaunchConfiguration> getAllConfigs() {
```
danrubel: What's the naming convention for things like getters?
Should we be using this pattern or `List<LaunchConfigurations> get allConfigs {` ?

danrubel:
LGTM

lib/flutter/flutter_launch.dart
devoncarew: This was mostly an artifact of the refactoring to get here; not all clients use it but I can make it a normal param in any case.

lib/flutter/flutter_launch.dart
devoncarew: yup! will remove

lib/launch/launch_configs.dart
devoncarew: I think `get` for things that are inexpensive to call, like O(1) amortized. This could be a `get`ter as well -

PR #875, opened 2016-03-17
Support custom build rules in BUILD files

BlackHC:
There are lots of cases where the markers are not finding anything because we only use wrapper rules. Just search for dart in the BUILD file.

### DIFF
--- a/lib/projects.dart
+++ b/lib/projects.dart
@@ -509,13 +509,11 @@ class ProjectScanJob extends Job {
 }
 
 bool _isDartBuildFile(File file) {
-  const String marker1 = '/dart/build_defs';
-  const String marker2 = 'dart_library(';
-  const String marker3 = 'dart_analyzed_library';
-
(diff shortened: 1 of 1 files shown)


devoncarew:
That does make it easy! I can land as is, but do you think false positives will be an issue? Would a regex that looked for `dart` as it's own word - that could match `/dart/`, `dart_`, `_dart_` - work for your case?

devoncarew:
@BlackHC, thanks for the PR. I'm going to close this in favor of https://github.com/dart-atom/dartlang/pull/912, which does some slightly more specific checks. We should have a new version out shortly.

BlackHC:
Sorry for the late reply! :( I guess the notification got filtered by inbox and I don't check github that often myself. Thanks for submitting and closing this!

devoncarew:
No worries! The patch is in the current shipping version. - hopefully working for your BUILD files.

PR #877, opened 2016-03-18
move create project functionality into dartlang

danrubel:
* move create project functionality into dartlang
* revise it to use the new dartino create project cmdline utility

@devoncarew

### DIFF
--- a/lib/plugin.dart
+++ b/lib/plugin.dart
@@ -193,6 +193,7 @@ class AtomDartPackage extends AtomPackage {
       });
     });
     _addCmd('atom-workspace', 'dartlang:send-feedback', (_) => _handleSendFeedback());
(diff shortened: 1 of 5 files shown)


lib/plugin.dart
```diff
@@ -193,6 +193,7 @@ class AtomDartPackage extends AtomPackage {
       });
     });
     _addCmd('atom-workspace', 'dartlang:send-feedback', (_) => _handleSendFeedback());
+    _addCmd('atom-workspace', 'dartino:create-new-proj', dartino.createNewProject);
```
devoncarew: `dartino:create-new-proj` ==> `dartino:create-new-project`?

devoncarew:
lgtm

lib/plugin.dart
danrubel: Good point. Will address in next PR.

PR #878, opened 2016-03-18
remove 30-40 strong mode issues

devoncarew:
- re-analyze with the latest strong mode and remove 30-40 issues found. Mostly just making strong mode happy, though it did uncover two real issues with method return types

@danrubel


danrubel:
LGTM

Issue #880, opened 2016-03-18
Uncaught NullError: method not found: 'get$isEmpty' on null


rayk:
[Enter steps to reproduce below:]

1. ...
2. ...

**Atom Version**: 1.6.0
**System**: Mac OS X 10.11.3
**Thrown From**: [dartlang](https://github.com/dart-atom/dartlang) package, v0.6.6

### Stack Trace

Uncaught NullError: method not found: 'get$isEmpty' on null

```
At /Users/rayk/.atom/packages/dartlang/web/entry.dart.js:26631

TypeError: Cannot read property 'get$isEmpty' of null
    at [object Object].J.get$isEmpty$asx (/Users/rayk/.atom/packages/dartlang/web/entry.dart.js:74167:42)
    at FlowControlSection.dart.FlowControlSection._handleIsolateChange$1 (/Users/rayk/.atom/packages/dartlang/web/entry.dart.js:35484:18)
    at [object Object].<anonymous> (/Users/rayk/.atom/packages/dartlang/web/entry.dart.js:27944:32)
    at FocusManager.dart.FocusManager._notifyIsolateListeners$0 (/Users/rayk/.atom/packages/dartlang/web/entry.dart.js:35888:18)
    at DebuggerView.dart.DebuggerView._handleIsolateResumed$1 (/Users/rayk/.atom/packages/dartlang/web/entry.dart.js:35141:14)
    at [object Object].<anonymous> (/Users/rayk/.atom/packages/dartlang/web/entry.dart.js:27944:32)
    at _RootZone.dart._RootZone.runUnaryGuarded$2 (/Users/rayk/.atom/packages/dartlang/web/entry.dart.js:49713:20)
    at _BroadcastSubscription.dart._BufferingStreamSubscription._sendData$1 (/Users/rayk/.atom/packages/dartlang/web/entry.dart.js:48958:20)
    at _DelayedData.dart._DelayedData.perform$1 (/Users/rayk/.atom/packages/dartlang/web/entry.dart.js:49104:18)
    at _PendingEvents_schedule_closure.dart._PendingEvents_schedule_closure.call$0 (/Users/rayk/.atom/packages/dartlang/web/entry.dart.js:49157:16)
    at [object Object].dart._microtaskLoop (/Users/rayk/.atom/packages/dartlang/web/entry.dart.js:46925:27)
[...]

PR #881, opened 2016-03-18
fix an npe in the stack frame view

devoncarew:
- fix an npe in the stack frame view (fix https://github.com/dart-atom/dartlang/issues/880)

@danrubel


danrubel:
LGTM

PR #882, opened 2016-03-18
Fix grammar rule for triple slash doc comments

jacehensley-wf:
Allows for doc comments to not have any content. Useful when a doc comment has a new line in order to make a new paragraph.

Came across this when setting custom CSS for doc comments.


devoncarew:
lgtm, thanks! (it looks like for the themes I've been using, the styles for `line` and `block` comments are the same)

PR #883, opened 2016-03-18
open some editors in the 'pending' mode

devoncarew:
Adjust to a change in atom apis; open many editors in a 'pending' mode. This will prevent a proliferation of new editors when traversing through files.

@danrubel


danrubel:
LGTM

PR #887, opened 2016-03-21
Complete ctors

devoncarew:
- refactor the location of the code completion files
- fix a bug in code completion where the class name was not shown when completing named constructors
- show icons in completions when completing suggestions for material design icons

@danrubel 

screen shot 2016-03-20 at 10 48 45 pm


danrubel:
LGTM

Issue #885, opened 2016-03-19
keyboard focus not moved when selecting "rename refactoring"


skybrian:
To reproduce:
  - right-click on a class and select "Rename".
  - start typing the new name
  - Instead of editing the new name in the popup, it edits the original file.

Workaround:
  -  use the mouse to click in the popup

Atom 1.5.4, dartlang 0.6.6

Issue #151, opened 2015-08-01
implement a `find type` dialog


devoncarew:

sethladd:
Use case: "I heard of this thing called Scaffold. What is it?" -- prompted us to ask "how would I ask Atom "show me the Scaffold thing"

PR #888, opened 2016-03-21
implement a find type dialog

devoncarew:
- implement a find type dialog (fix https://github.com/dart-atom/dartlang/issues/151)
- fix a keyboard focus issue with the rename dialog (fix https://github.com/dart-atom/dartlang/issues/885)

@danrubel

### DIFF
--- a/lib/atom_utils.dart
+++ b/lib/atom_utils.dart
@@ -10,6 +10,7 @@ import 'dart:html' show DivElement, Element, HttpRequest, Node, NodeValidator,
     NodeTreeSanitizer, window;
 import 'dart:js';
 
+import 'package:atom/src/js.dart';
 import 'package:atom/node/node.dart';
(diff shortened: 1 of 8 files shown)


lib/atom_utils.dart
```diff
@@ -76,6 +71,15 @@ Future<String> promptUser(String prompt,
 
   Panel panel = atom.workspace.addModalPanel(item: element, visible: true);
 
+  Timer.run(() {
+    // There's a bug in Dart's JS interop where we can't see custom elements
+    // well. In order to work around it, we get the raw JS object, wrap it in a
+    // JsObject, pass that to a semantic wrapper around TextEditorElement, and
+    // then call `focused()`.
```
danrubel: This had me stumped as to why promptUser was not getting focus recently. Thanks for fixing this!

danrubel:
LGTM

Issue #892, opened 2016-03-22
Uncaught NullError: method not found: 'apply' on null


sethladd:
[Enter steps to reproduce below:]

1. Cmd-Shift-P
1. type "find type"
1. Enter `StatefulWidget`

**Atom Version**: 1.6.0
**System**: Mac OS X 10.11.3
**Thrown From**: [dartlang](https://github.com/dart-atom/dartlang) package, v0.6.7

### Stack Trace

Uncaught NullError: method not found: 'apply' on null

```
At /Users/sethladd/.atom/packages/dartlang/web/entry.dart.js:26802

TypeError: Cannot read property 'apply' of undefined
    at JsObject.dart.JsObject.callMethod$2 (/Users/sethladd/.atom/packages/dartlang/web/entry.dart.js:61224:43)
    at JsObject.dart.JsObject.callMethod$1 (/Users/sethladd/.atom/packages/dartlang/web/entry.dart.js:61227:21)
    at TextEditor.dart.ProxyHolder.invoke$4 (/Users/sethladd/.atom/packages/dartlang/web/entry.dart.js:64819:23)
    at TextEditor.dart.ProxyHolder.invoke$1 (/Users/sethladd/.atom/packages/dartlang/web/entry.dart.js:64826:21)
    at FindTypeHelper__handleFindType_closure.dart.FindTypeHelper__handleFindType_closure.call$1 (/Users/sethladd/.atom/packages/dartlang/web/entry.dart.js:63477:52)
    at _RootZone.dart._RootZone.runUnary$2 (/Users/sethladd/.atom/packages/dartlang/web/entry.dart.js:49821:20)
    at _Future__propagateToListeners_handleValueCallback.dart._Future__propagateToListeners_handleValueCallback.call$0 (/Users/sethladd/.atom/packages/dartlang/web/entry.dart.js:48164:47)
    at [object Object].dart._Future.static._Future__propagateToListeners (/Users/sethladd/.atom/packages/dartlang/web/entry.dart.js:48072:120)
    at _Future.dart._Future._completeWithValue$1 (/Users/sethladd/.atom/packages/dartlang/web/entry.dart.js:47931:11)
    at _Future__asyncComplete_closure0.dart._Future__asyncComplete_closure0.call$0 (/Users/sethladd/.atom/packages/dartlang/web/entry.dart.js:48149:20)
[...]

Issue #891, opened 2016-03-22
HTML is displayed when code completing Icons data


sethladd:
Not sure if we intended to display HTML here (when code completing Icons)

![screen shot 2016-03-22 at 9 30 47 am](https://cloud.githubusercontent.com/assets/5479/13959303/d9ca902e-f010-11e5-9686-3a246e265347.png)

Atom 1.6.0, dartlang 0.6.7, SDK 1.15.0-dev.5.0 running on macos.

PR #893, opened 2016-03-22
support case insensitive searching in the find type dialog

devoncarew:
- fix https://github.com/dart-atom/dartlang/issues/892, fix https://github.com/dart-atom/dartlang/issues/891
- strip html markup from the code completion results
- support case insensitive searching in the find type dialog

@danrubel


devoncarew:
(TBR to fix the exception)

danrubel:
LGTM

PR #899, opened 2016-03-24
remove some keybindings

devoncarew:
fix https://github.com/dart-atom/dartlang/issues/870

@danrubel, w/ this change we'll need to run the rebuild command through the command pallette - `rebuild restart dev`.


danrubel:
LGTM. I'll add them to my local key bindings.

PR #910, opened 2016-03-28
type dialog work

devoncarew:
Some refactoring to the type dialog to make it easier to add completions in the dialog.
- instead of a stream of results, use a Future
- if no results, beep; if 1, jump to that result; if 2 or more show a results view
- some work to support overriding the completions handler for a text editor

@danrubel


danrubel:
LGTM

PR #911, opened 2016-03-28
move show SDK docs into dartlang

danrubel:
move show SDK docs into dartlang so that duplicate code can be removed from the dartino plugin.
@devoncarew


devoncarew:
lgtm

PR #912, opened 2016-03-28
improve our support for BUILD files

devoncarew:
Widen what we look for in BUILD files; any `dart` reference (surrounded by non-word chars). 

@danrubel


danrubel:
LGTM

PR #919, opened 2016-03-31
refactor dartlang package functionality to mirror atom.dart

danrubel:
This is the second step towards having flutter, dartino, and dartlang using the same package infrastructure.
@devoncarew


devoncarew:
> This is the second step towards having flutter, dartino, and dartlang using the same package infrastructure.

:+1: lgtm

PR #923, opened 2016-04-04
remove workspace and config classes from dartlang

danrubel:
remove workspace and config classes from dartlang
and use those same classes in atom.dart
@devoncarew


devoncarew:
lgtm

Issue #931, opened 2016-04-11
remove auto-format on save


devoncarew:
It's buggy and interacts badly with auto-save. Re-add it when the implementation is more complete.

pylaligand:
Did this become available with version 0.6.12 of dartlang? No formatting happens on save since I updated earlier today. What's the alternative, dartfmt? I have to say auto-format on save was an awesome feature, it's super annoying to not have it anymore (especially as you could disable it if you wanted to - and the option is still available in the package settings).

devoncarew:
I had too many bug reports about this feature and didn't the bandwidth to fix it. Totally agree that it's annoying to have it gone, esp. if you were using it.

The option is still visible to you as you had set it to the non-default value, and the value is still in your config.cson file. Users who hadn't changed the setting won't see it.

I'm open to ideas here, but I'm having to prioritize work given limited resources (me!), and turning off buggy features reduces my workload :)

pulyaevskiy:
@devoncarew could you provide a list of all critical issues with auto-format on save, so we can try and brainstorm some ideas in this thread?

One option to consider: return the option but put "experimental/buggy/use-at-your-own-risk" label on it (chose one)?

devoncarew:
These two are the main issues:

- https://github.com/dart-atom/dartlang/issues/864 (race condition between format on save and some refactorings)
- https://github.com/dart-atom/dartlang/issues/927 (format on save doesn't work well with atom's auto-save feature)

Adding an `[experimental]` tag to the UI option would work.

pulyaevskiy:
Thanks for the #963 ! I've been sitting on the old version because this option is very useful.

PR #935, opened 2016-04-11
remove the auto-format option

devoncarew:
- remove the auto-format on save option - fix https://github.com/dart-atom/dartlang/issues/931

@danrubel


danrubel:
LGTM

Issue #924, opened 2016-04-05
Uncaught [Analyzer RequestError method: edit.sortMembers, code: SORT_MEMBERS_PARSE_ERRORS, message: Error during edit.sortMembers: file has 7 scan/parse errors.]


mezoni:
[Enter steps to reproduce below:]

1. ...
2. ...

**Atom Version**: 1.5.4
**System**: Ubuntu 15.10
**Thrown From**: [dartlang](https://github.com/dart-atom/dartlang) package, v0.6.9

### Stack Trace

Uncaught [Analyzer RequestError method: edit.sortMembers, code: SORT_MEMBERS_PARSE_ERRORS, message: Error during `edit.sortMembers`: file has 7 scan/parse errors.]

```
At /home/antonina/.atom/packages/dartlang/web/entry.dart.js:26851

[Analyzer RequestError method: edit.sortMembers, code: SORT_MEMBERS_PARSE_ERRORS, message: Error during `edit.sortMembers`: file has 7 scan/parse errors.]
    at [object Object].dart.wrapException (/home/antonina/.atom/packages/dartlang/web/entry.dart.js:27916:17)
    at _rootHandleUncaughtError_closure.dart._rootHandleUncaughtError_closure.call$0 (/home/antonina/.atom/packages/dartlang/web/entry.dart.js:49652:19)
    at [object Object].dart._microtaskLoop (/home/antonina/.atom/packages/dartlang/web/entry.dart.js:46895:27)
    at [object Object].dart._startMicrotaskLoop (/home/antonina/.atom/packages/dartlang/web/entry.dart.js:46901:11)
    at TimerImpl_internalCallback0.dart.TimerImpl_internalCallback0.call$0 (/home/antonina/.atom/packages/dartlang/web/entry.dart.js:27231:23)
    at invokeClosure_closure.dart.invokeClosure_closure.call$0 (/home/antonina/.atom/packages/dartlang/web/entry.dart.js:29099:29)
    at _IsolateContext.dart._IsolateContext.eval$1 (/home/antonina/.atom/packages/dartlang/web/entry.dart.js:26842:25)
    at [object Object].dart._callInIsolate (/home/antonina/.atom/packages/dartlang/web/entry.dart.js:26484:28)
    at dart.invokeClosure (/home/antonina/.atom/packages/dartlang/web/entry.dart.js:28064:20)
    at /home/antonina/.atom/packages/dartlang/web/entry.dart.js:28085:18
```

### Commands

```
[...]

PR #936, opened 2016-04-11
display errors from the analysis server better

devoncarew:
- display errors from the analysis server better (wrap longer calls to the analysis server in `AnalysisRequestJob`s)
- fix https://github.com/dart-atom/dartlang/issues/924

@danrubel


danrubel:
LGTM

PR #937, opened 2016-04-13
Increase the recursion depth searching for projects

ianloic:
We need this for our project. I kept missing things in a sub-sub-sub-sub folder when I was doing refactorings because there were dart projects that weren't getting analyzed.

Actually I wonder how expensive it would be just to search all the way down. For the most part it's just reading directories. Is the concern mostly about a hypothetical giant mapped network drive containing billions of lines of source code? Could we just special-case that one?

### DIFF
--- a/lib/projects.dart
+++ b/lib/projects.dart
@@ -38,7 +38,7 @@ String getWorkspaceRelativeDescription(String path) {
 /// A class to locate Dart projects in Atom and listen for new or removed Dart
 /// projects.
 class ProjectManager implements Disposable, ContextMenuContributor {
-  static const int _recurseDepth = 3;
+  static const int _recurseDepth = 4;
 
   /// Return whether the given directory is a Dart project.
   static bool isDartProject(Directory dir) {


devoncarew:
> Is the concern mostly about a hypothetical giant mapped network drive

Yeah, that's pretty much it. We want to be careful to avoid the pathological cases - I have run into problems w/ this on other tools.

I'm fine w/ special casing them, to avoid the most common pathological situations. I will probably need to tweak this PR to avoid one or two that I know about.

ianloic:
Well this CL doesn't remove the depth, it just bumps it from 3 to 4. The challenge is that as soon as someone has a dart project 5 directories deep they'll just be surprised that it doesn't work :(

devoncarew:
There's one project I know of that will start seeing projects they don't want analyzed w/ this bump :) So I want to special case that.

Ideally we'd be able to use something like an `.analysis_options` file to hide certain subdirs, but those are currently only read for directories with pubspecs.

ianloic:
I have a CL that I'm working on to exclude directories, because I get a bunch of crap from out/ and third_party/ in my project. I'm doing it in a global pref because I don't know a better way. Does Atom have the concept of directory-specific prefs or is that something we'd have to build ourselves?

devoncarew:
Sorry for the delay! Re: prefs, it's something we'd have to build. The atom config system works globally (~/.atom/config.cson). Their per session state system saves info based on the directory you opened atom in. It's a map that gets restored to you on plugin startup (https://github.com/dart-atom/dartlang/blob/master/lib/plugin.dart#L117).

lgtm

PR #938, opened 2016-04-13
Add an option to exclude particular directories. By default exclude .git

ianloic:
In our project we have a third_party/ directory that contains a lot of code that we don't maintain. Our build system also puts a lot of stuff in a out/ directory - some of which look like dart packages but aren't well formed. We don't want to treat things that look like dart projects in there as dart projects in the editor.

This may actually belong in a directory-specific dart-atom configuration, but I'm not sure what the best way to add that is.

### DIFF
--- a/lib/plugin.dart
+++ b/lib/plugin.dart
@@ -391,6 +391,14 @@ class AtomDartPackage extends AtomPackage {
         'type': 'boolean',
         'default': false,
         'order': 13
+      },
+
+      'excludeDirectories': {
+        'title': '[Experimental] Exclude directories from indexing',
+        'description': 'A comma separated list of directory names to avoid '
+        'indexing',
+        'type': 'string',
+        'default': '.git'
       }
     };
   }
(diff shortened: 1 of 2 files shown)


devoncarew:
@ianloic, I'm thinking about using a marker file - possibly called `.dartignore` - in a directory to tell me not to scan down into it for dart projects. Would that work for your use case? It would probably need to be committed to your repo to be useful.

Related discussion here: https://github.com/flutter/flutter/pull/3368#discussion_r59960349.

devoncarew:
see https://github.com/dart-atom/dartlang/pull/943

ianloic:
I was thinking about that but it would mean checking in (and not rm -fr
ing) our out/ directory. If you implement it I'll use it :)

devoncarew:
Yeah, it would be more flexible if we supported a glob based ignore file, but I'm hesitant to do all that work when search for dart projects (reading files and doing glob matching). The marker file is a pretty simple solution.

I can land this - a pref to ignore named directories. But it would be a setting that each user of the project would need to set. I would probably remove the UI portion - the definition in plugin.dart. You'd still be able to set and read from that value, it just wouldn't have a UI if there was no value.

ianloic:
I think dropping this change makes sense. This should be per repo not per
user.

devoncarew:
sgtm

PR #957, opened 2016-04-20
add support for isolate reloading

devoncarew:
- add experimental support for isolate reloading

screen shot 2016-04-20 at 10 28 35 am

@danrubel


lib/debug/observatory_debugger.dart
```diff
@@ -287,6 +290,10 @@ class ObservatoryConnection extends DebugConnection {
       case EventKind.kIsolateExit:
         _handleIsolateDeath(ref);
         break;
+      case 'IsolateReload':
```
danrubel: Is there a constant for this?

devoncarew: Not yet - it's not part of the service protocol spec (but is supported by some VMs on tip-of-tree).

danrubel:
LGTM

PR #962, opened 2016-04-22
more strong mode fixes

devoncarew:
- add additional type info to fix strong mode issues
- down from 62 issues to 20

@danrubel


danrubel:
LGTM

PR #963, opened 2016-04-22
re-enable format on save

devoncarew:
Re: https://github.com/dart-atom/dartlang/issues/931, re-enable format on save, with some experimental messaging.

screen shot 2016-04-22 at 8 26 40 am

@danrubel


danrubel:
LGTM

PR #965, opened 2016-04-23
save dirty editors before running pub get/upgrade

devoncarew:
- save dirty editors before running pub get/upgrade (fix https://github.com/flutter/flutter/issues/2300)

@danrubel


danrubel:
LGTM

PR #966, opened 2016-04-25
use Dart SDK that ships with Dartino SDK

danrubel:
Fixes https://github.com/dartino/sdk/issues/452
@devoncarew


devoncarew:
lgtm

Issue #967, opened 2016-04-25
New code completion not appearing in Atom


jwren:
Fix for Server here https://github.com/dart-lang/sdk/issues/24254 is not showing up in Atom

Atom 1.1.0, dartlang 0.6.9, SDK 1.16.0-edge.8bcb84b9871dfcb2bf1c92baf6a6cbaad87c1fea running on linux.

new CompletionSuggestion(CompletionSuggestionKind.IDENTIFIER,
      DART_RELEVANCE_DEFAULT, name, name.length, 0, false, false);

CompletionSuggestion(CompletionSuggestionKind kind, int relevance, String completion, int selectionOffset, int selectionLength, bool isDeprecated, bool isPotential, ...)

PR #968, opened 2016-04-25
fix to code completion

devoncarew:
- fix an npe in code completion (fix https://github.com/dart-atom/dartlang/issues/967)

@jwren


jwren:
LGTM

Issue #970, opened 2016-04-30
sdk location code needs to be unified


devoncarew:
We have logic for this in three separate places, and some race conditions as a result. We should unify the code and ensure that it's deterministic.

PR #974, opened 2016-05-02
fix sdk auto-location

devoncarew:
- fix an issue where the dartino plugin could over-write a valid dart sdk setting
- ensure that he flutter code doesn't run if the flutter plugin isn't installed
- fix https://github.com/dart-atom/dartlang/issues/970

@pq @danrubel


danrubel:
LGTM

lib/dartino/dartino_util.dart
```diff
@@ -137,13 +137,12 @@ class _Dartino {
       var sdk = sdkFor(null);
       if (sdk != null && sdk.validate()) {
 
-        // Given a valid Dartino SDK, now validate the Dart SDK
-        if (sdkManager.sdk?.isValidSdk != true) {
+        // If there's no Dart SDK configured, use the one provided by Dartino.
+        if (sdkManager.noSdkPathConfigured) {
```
danrubel: If the Dart SDK is not valid (e.g. path is set but directory has been removed) they wouldn't you want the sdkPath set?

devoncarew: This was causing issues on Phil's machine. I think the ideal solution might be to notify the user if we see an invalid sdk path? And give them an opportunity to correct it. We may want to unify the sdk location logic for dartlang, flutter, and dartino to make things a bit more deterministic. I suspect the order of things now depends on the timings of various Futures.

lib/flutter/flutter_sdk.dart
```diff
@@ -44,15 +46,17 @@ class FlutterSdkManager implements Disposable {
...
-    _disposables.add(atom.commands.add('atom-workspace', 'flutter:auto-locate-flutter-sdk', (_) {
-      new SdkLocationJob(this).schedule();
-    }));
-    _disposables.add(atom.commands.add('atom-workspace', 'flutter:show-flutter-sdk-info', (_) {
-      showInstallationInfo();
-    }));
-    _disposables.add(atom.commands.add('atom-workspace', 'flutter:version', (_) {
-      showVersionInfo();
-    }));
+    if (Flutter.hasFlutterPlugin()) {
```
danrubel: Is there any issue have the commands below defined if the flutter plugin is not installed?

devoncarew: Not really, it's just that they are visible from the command palette, which will be odd, when they don't have the flutter plugin installed.

pq:
LGTM.

Thanks!

lib/flutter/flutter_sdk.dart
danrubel: Good point. I will do something similar with the dartino specific commands so that they show up only if the dartino plugin is installed.

danrubel: Landed PR to [show Dartino specific commands iff dartino plugin installed](https://github.com/dart-atom/dartlang/pull/984/commits/d1e3ffa5b01617a3bad808e90e8d53f069839d4f)

Issue #969, opened 2016-04-29
Uncaught NullError: method not found: 'matchAsPrefix$2' on null


mezoni:
[Enter steps to reproduce below:]

1. ...
2. ...

**Atom Version**: 1.7.2
**System**: Ubuntu 16.04
**Thrown From**: [dartlang](https://github.com/dart-atom/dartlang) package, v0.6.13

### Stack Trace

Uncaught NullError: method not found: 'matchAsPrefix$2' on null

```
At /home/andrew/.atom/packages/dartlang/web/entry.dart.js:26807

TypeError: Cannot read property 'matchAsPrefix$2' of null
    at [object Object].J.matchAsPrefix$2$s (/home/andrew/.atom/packages/dartlang/web/entry.dart.js:75766:40)
    at JSString.dart.JSString.startsWith$2 (/home/andrew/.atom/packages/dartlang/web/entry.dart.js:26156:18)
    at JSString.dart.JSString.startsWith$1 (/home/andrew/.atom/packages/dartlang/web/entry.dart.js:26159:21)
    at [object Object].J.startsWith$1$s (/home/andrew/.atom/packages/dartlang/web/entry.dart.js:75853:41)
    at UriResolver__resolvePathToUris_closure0.dart.UriResolver__resolvePathToUris_closure0.call$1 (/home/andrew/.atom/packages/dartlang/web/entry.dart.js:34042:15)
    at _RootZone.dart._RootZone.runUnary$2 (/home/andrew/.atom/packages/dartlang/web/entry.dart.js:49135:20)
    at _Future__propagateToListeners_handleValueCallback.dart._Future__propagateToListeners_handleValueCallback.call$0 (/home/andrew/.atom/packages/dartlang/web/entry.dart.js:47478:47)
    at [object Object].dart._Future.static._Future__propagateToListeners (/home/andrew/.atom/packages/dartlang/web/entry.dart.js:47386:120)
    at _Future.dart._Future._completeWithValue$1 (/home/andrew/.atom/packages/dartlang/web/entry.dart.js:47245:11)
    at _Future__asyncComplete_closure0.dart._Future__asyncComplete_closure0.call$0 (/home/andrew/.atom/packages/dartlang/web/entry.dart.js:47463:20)
    at [object Object].dart._microtaskLoop (/home/andrew/.atom/packages/dartlang/web/entry.dart.js:46300:27)
[...]

PR #975, opened 2016-05-02
fix an npe in the debugger

devoncarew:
Fix an NPE in the debugger when no self ref package is found (fix https://github.com/dart-atom/dartlang/issues/969).

@danrubel


danrubel:
LGTM

PR #988, opened 2016-05-15
refactor tests

devoncarew:
@danrubel, some refactoring so that the dartino tests can be run from the command-line, and run them on the bots as part of the `all.dart` script.


danrubel:
LGTM

PR #990, opened 2016-05-18
don't pass in --checked

devoncarew:
@danrubel, no longer pass `--checked` into the flutter cli tool - that arg has been removed.

(TBR)


danrubel:
LGTM

PR #995, opened 2016-05-20
drop frame numbers from the debugger ui

devoncarew:
- drop frame numbers from the debugger ui (fix https://github.com/flutter/flutter/issues/4062)

@danrubel 

screen shot 2016-05-19 at 9 19 20 pm

### DIFF
--- a/lib/debug/debugger_ui.dart
+++ b/lib/debug/debugger_ui.dart
@@ -492,8 +492,7 @@ class ExecutionTab extends MTab {
           text: locationText,
           c: 'debugger-secondary-info right-aligned overflow-hidden-ellipsis'
         )..flex()
-      ])..flex(),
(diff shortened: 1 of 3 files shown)


danrubel:
LGTM

jamesr:
Without frame numbers it's really hard to communicate about a callstack without saying really awkward stuff like "look at the members in the stack frame for the second frame underscore flutter demo state dot underscore increment counter".

devoncarew:
I'm hoping the number of the frame isn't critical to helping users debug an issue - that as they're stepping into and out of methods, it's only the top few frames that are important to their current debug session.

I'd mostly added the numbers in there in the first place to give people a sense for how deep in the call stack they were. But they used up valuable screen real estate, and were confusing wrt the console stack numbers.

Issue #996, opened 2016-05-20
Uncaught NullError: method not found: 'get$suspended' on null


mezoni:
[Enter steps to reproduce below:]

1. ...
2. ...

**Atom Version**: 1.7.3
**System**: Ubuntu 16.04
**Thrown From**: [dartlang](https://github.com/dart-atom/dartlang) package, v0.6.15

### Stack Trace

Uncaught NullError: method not found: 'get$suspended' on null

```
At /home/user/.atom/packages/dartlang/web/entry.dart.js:26765

TypeError: Cannot read property 'get$suspended' of null
    at FlowControlSection.dart.FlowControlSection._pauseResume$0 (/home/user/.atom/packages/dartlang/web/entry.dart.js:34364:24)
    at [object Object].<anonymous> (/home/user/.atom/packages/dartlang/web/entry.dart.js:28072:32)
    at CoreElement_click_closure.dart.CoreElement_click_closure.call$1 (/home/user/.atom/packages/dartlang/web/entry.dart.js:35538:21)
    at _RootZone.dart._RootZone.runUnaryGuarded$2 (/home/user/.atom/packages/dartlang/web/entry.dart.js:48183:20)
    at _ForwardingStreamSubscription.dart._BufferingStreamSubscription._sendData$1 (/home/user/.atom/packages/dartlang/web/entry.dart.js:47428:20)
    at _ForwardingStreamSubscription.dart._BufferingStreamSubscription._async$_add$1 (/home/user/.atom/packages/dartlang/web/entry.dart.js:47378:16)
    at _ForwardingStreamSubscription.dart._ForwardingStreamSubscription._async$_add$1 (/home/user/.atom/packages/dartlang/web/entry.dart.js:47845:14)
    at [object Object].J._async$_add$1$x (/home/user/.atom/packages/dartlang/web/entry.dart.js:75512:41)
    at _WhereStream.dart._WhereStream._handleData$2 (/home/user/.atom/packages/dartlang/web/entry.dart.js:47922:13)
    at _ForwardingStreamSubscription.dart._ForwardingStreamSubscription._handleData$1.H.computeSignature.T.func [as _handleData$1] (/home/user/.atom/packages/dartlang/web/entry.dart.js:47873:22)
    at [object Object].<anonymous> (/home/user/.atom/packages/dartlang/web/entry.dart.js:28078:32)
[...]

PR #997, opened 2016-05-20
Show exception

devoncarew:
- when stopping on an exception, display the toString() of an exception in the console (fix https://github.com/flutter/flutter/issues/4037)
- when stopping on an exception, display the exception object in the stack frame view of the first frame
- fix an npe when hitting the resume button (fix https://github.com/dart-atom/dartlang/issues/996)

@danrubel 

screen shot 2016-05-19 at 10 51 46 pm

screen shot 2016-05-20 at 8 01 48 am


devoncarew:
(also, change the break-on-exceptions default to `uncaught` from `all`. We were too aggressive about stopping on flutter exceptions: https://github.com/flutter/flutter/issues/3899. We'll need to coordinate with the flutter framework to show exception locations for handled exceptions)

lib/debug/observatory_debugger.dart
```diff
@@ -398,8 +395,38 @@ class ObservatoryConnection extends DebugConnection {
...
+    return ExceptionPauseMode.kUnhandled;
+  }
+
+  void _printExceptionToConsole(ObservatoryIsolate isolate, InstanceRef exception) {
+    if (exception.kind == InstanceKind.kString) {
+      launch.pipeStdio(
+        "exception: '${exception.valueAsString}'\n",
+        error: true
+      );
+    } else if (exception.valueAsString != null) {
```
danrubel: Combine if(A) and else if(B) into if(A||B) because blocks are identical ?

danrubel:
LGTM

lib/debug/observatory_debugger.dart
devoncarew: will do

PR #998, opened 2016-05-21
Dartino debugging on device

danrubel:
First cut debugging support - not enabled by default
@devoncarew

### DIFF
--- a/lib/dartino/launch_dartino.dart
+++ b/lib/dartino/launch_dartino.dart
@@ -3,10 +3,12 @@ import 'dart:async';
 import 'package:atom/node/process.dart';
 import 'package:logging/logging.dart';
 
+import '../debug/observatory_debugger.dart';
 import '../launch/launch.dart';
 import '../projects.dart';
 import '../state.dart';
 import 'dartino.dart';
+import 'sdk/dartino_sdk.dart';
 import 'sdk/sdk.dart';
 
 final Logger _logger = new Logger('atom.dartino_launch');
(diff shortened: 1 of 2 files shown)


lib/dartino/launch_dartino.dart
```diff
@@ -113,4 +115,56 @@ class DartinoLaunch extends Launch {
...
+    runner.onStdout.listen((str) {
+      if (str.startsWith('localhost:')) {
+        try {
+          portCompleter.complete(int.parse(str.substring(10)));
+        } catch (e) {
+          pipeStdio('Failed to parse observatory port from "$str"',
+              error: true);
+        }
+      }
+      pipeStdio(str, subtle: true);
```
devoncarew: You might want to move this above the if statement.

lib/dartino/launch_dartino.dart
```diff
@@ -113,4 +115,56 @@ class DartinoLaunch extends Launch {
...
+    });
+    runner.onStderr.listen((str) => pipeStdio('\n$str\n', error: true));
+    runner.execStreaming().then((int exitCode) {
+      pipeStdio('debug session exit code is $exitCode', highlight: true);
+      launchTerminated(exitCode, quiet: true);
+    });
+    int observatoryPort = await portCompleter.future
+        .timeout(new Duration(seconds: 5), onTimeout: () => null);
```
devoncarew: Does a 5 sec timeout work for most dartino launches?

devoncarew:
Lgtm! Exciting to see it hooked it.

lib/dartino/launch_dartino.dart
danrubel: Good point. Done.

lib/dartino/launch_dartino.dart
danrubel: Don't know yet. I picked 5 seconds as a way to exit out of the process if the port is not found in a reasonable amount of time, but perhaps I could ask the user if they want to continue waiting or ? I'll add a note here to ponder and follow up with a better solution.

danrubel:
Merged https://github.com/dart-atom/dartlang/commit/6459d053b05cd2fc30078415951180a701ef9518

lib/dartino/launch_dartino.dart
devoncarew: Most flutter apps would start in less than 5 secs, but the outliers - ones  that start but start slowly - are more in the neighborhood of 20 secs. No idea what dartino apps look like :) well probably get a sense of that just by using it.

PR #1000, opened 2016-05-23
Allow ~ in Dartino SDK path

danrubel:
fixes https://github.com/dartino/atom-dartino/issues/44
@devoncarew


devoncarew:
lgtm

We've had a similar bug report for the dart_sdk pref field.

danrubel:
> We've had a similar bug report for the dart_sdk pref field.

This should fix that... https://github.com/dart-atom/dartlang/pull/1002

PR #1008, opened 2016-05-27
change to using flutter run --resident

devoncarew:
-- change to using `flutter run --resident`; it's a long-running flutter launch, designed to speed up the edit/refresh dev. cycle

@danrubel


lib/flutter/flutter_launch.dart
```diff
@@ -148,78 +156,54 @@ class _LaunchInstance {
...
-      if (_device != null) {
-        args.add('--device-id');
-        args.add(_device.id);
-      }
-
-      ProcessRunner flutterStop = _flutter(flutter, args, cwd: project.path);
-      flutterStop.execSimple().catchError((e) {
-        _logger.finer('Error from flutter stop', e);
+      return new Future.delayed(new Duration(milliseconds: 250), () {
+        return _runner.kill();
```
danrubel: Will _runner ever be null here? Is there a race condition? Or is _runner never set to null once it is assigned?

danrubel:
LGTM

lib/flutter/flutter_launch.dart
devoncarew: I'll add a `?` operator and cleanup up the `_runner` lifecycle a bit.

devoncarew:
Having a bit of trouble with this locally - it's not killing the app on the device when the user hits stop. Will do some triaging -

Issue #1013, opened 2016-05-28
Uncaught Invalid argument: null


mezoni:
[Enter steps to reproduce below:]

1. ...
2. ...

**Atom Version**: 1.7.3
**System**: Ubuntu 16.04
**Thrown From**: [dartlang](https://github.com/dart-atom/dartlang) package, v0.6.21

### Stack Trace

Uncaught Invalid argument: null

```
At /home/user/.atom/packages/dartlang/web/entry.dart.js:26712

Invalid argument: null
    at [object Object].dart.wrapException (/home/user/.atom/packages/dartlang/web/entry.dart.js:27786:17)
    at [object Object].dart.throwExpression (/home/user/.atom/packages/dartlang/web/entry.dart.js:27799:15)
    at JSString.dart.JSString.contains$2 (/home/user/.atom/packages/dartlang/web/entry.dart.js:26225:13)
    at JSString.dart.JSString.contains$1 (/home/user/.atom/packages/dartlang/web/entry.dart.js:26231:21)
    at ConsoleView.dart.ConsoleView.ConsoleView$2 (/home/user/.atom/packages/dartlang/web/entry.dart.js:33261:18)
    at [object Object].dart.ConsoleView.static.ConsoleView$ (/home/user/.atom/packages/dartlang/web/entry.dart.js:33280:14)
    at ConsoleController.dart.ConsoleController._launchAdded$1 (/home/user/.atom/packages/dartlang/web/entry.dart.js:32986:18)
    at [object Object].<anonymous> (/home/user/.atom/packages/dartlang/web/entry.dart.js:28034:32)
    at _RootZone.dart._RootZone.runUnaryGuarded$2 (/home/user/.atom/packages/dartlang/web/entry.dart.js:48319:20)
    at _BroadcastSubscription.dart._BufferingStreamSubscription._sendData$1 (/home/user/.atom/packages/dartlang/web/entry.dart.js:47570:20)
    at _BroadcastSubscription.dart._BufferingStreamSubscription._async$_add$1 (/home/user/.atom/packages/dartlang/web/entry.dart.js:47520:16)
    at _SyncBroadcastStreamController__sendData_closure.dart._SyncBroadcastStreamController__sendData_closure.call$1 (/home/user/.atom/packages/dartlang/web/entry.dart.js:46113:22)
[...]

PR #1014, opened 2016-05-30
fix an npe from the console; tweak console UI

devoncarew:
- fixed an NPE when starting command-line apps (fix https://github.com/dart-atom/dartlang/issues/1013)
- made the console output take up slightly less vertical space (less padding for the header and footer, and have the header take up one line instead of 2

@danrubel


danrubel:
LGTM

PR #1018, opened 2016-06-03
fix completions for import/show statements

danrubel:
When suggesting completions for
```
import '/a/path/to/myfile.dart' show ^
```
we were correctly suggesting top level functions, but incorrectly inserting parameters for those top level functions when we should only be inserting the identifier for the top level function. This change prevents parameters from being inserted when only an identifier should be inserted.
@devoncarew

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # dartlang plugin changelog
 
+## unreleased
+- fix completions for import/show statements
+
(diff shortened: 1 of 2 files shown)


devoncarew:
lgtm, thanks!

danrubel:
Merged https://github.com/dart-atom/dartlang/commit/f1d403b08523483c84ae717555b9436037eca07c

PR #1019, opened 2016-06-04
show the information about the analysis server's obs. port

devoncarew:
- show the information about the analysis server's obs. port
- make the analysis server debug option visible on the settings page

screen shot 2016-06-04 at 3 18 34 pm

@danrubel (likely will TBR)


devoncarew:
(to help with https://github.com/flutter/flutter/issues/4384)

lib/analysis/analysis_server_lib.dart
```diff
@@ -81,14 +81,14 @@ class Server {
...
 
-    try {
-      _onReceive.add(message);
+    if (!message.startsWith('{')) {
+      _logger.warning('unknown message: ${message}');
+      return;
+    }
 
+    try {
       var json = JSON.decode(message);
```
danrubel: try/catch around JSON.decode ?

lib/analysis_server.dart
```diff
@@ -126,6 +126,11 @@ class AnalysisServer implements Disposable {
     });
 
     onReceive.listen((String message) {
+      if (message.startsWith('Observatory listening') ||
+          message.startsWith('Observatory no longer listening')) {
+        atom.notifications.addInfo('Analysis server', detail: message.trim(), dismissable: true);
```
danrubel: Remove ```dismissable: true``` so that message automatically is removed after a while?

danrubel:
LGTM

lib/analysis_server.dart
devoncarew: This message will contain the port to connect to for the observatory connection, so we'll want it to persist for a while.

PR #1020, opened 2016-06-04
fix an issue with newly created launches

devoncarew:
Fix an issue - newly created launches would fail with the message `no handler for launch type ‘null’ found`. We were trying to parse the file before it had been created on disk.

@danrubel


danrubel:
Thank you for tracking that one down. LGTM!

PR #1024, opened 2016-06-07
add support for flutter run/build modes

devoncarew:
Add support for running flutter apps in the three build modes (debug, profile, release) (fix https://github.com/flutter/flutter/issues/4042).

screen shot 2016-06-06 at 4 27 41 pm

@danrubel

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # dartlang plugin changelog
 
+## unreleased
+- added the ability to run flutter apps in debug / profile / release modes
+
 ## 0.6.24
 - fix an issue with newly created launches
 - remove auto-resume during dartino debug startup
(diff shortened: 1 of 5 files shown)


danrubel:
LGTM

sethladd:
Nice! how do I control this in my YAML launch config?

devoncarew:
It's controlled through the toolbar GUI. Thinking about it, the state seemed more transient - per user, and not something you'd necessarily check in to scm.

devoncarew:
It's available in the latest release to try out -

sethladd:
> not something you'd necessarily check in to scm.

True. Should we remove `debug: true` from main.yaml then? Also, I thought we were ignoring `.atom` directory anyway? I always assumed that file shouldn't be checked in.

devoncarew:
Yup, re: removing debug: true.

We do exclude the .atom dir for the flutter project we create, but in general I think launch configs should be sharable between users. They should describe the application to launch - the type of app, the path to the main file, ... . Other launch settings, like debug/release, are more ephemeral.

PR #1025, opened 2016-06-07
split the mojo launch config out from the flutter one

devoncarew:
Split the mojo launch config out from the flutter one. The flutter one is already more specialized than the mojo one, and future work will make it more so.

@danrubel


danrubel:
LGTM

PR #1027, opened 2016-06-08
fix issues with remote debugger connect

devoncarew:
Fix issues with remote debugger connect:
- close the web socket connection when terminating debugging
- check to make sure there's a device available before trying to connect
- pipe embedder stdout to the console view

@danrubel


danrubel:
LGTM

PR #1033, opened 2016-06-10
switch over to using the daemon server's app.start to run flutter apps

devoncarew:
- switch over flutter app launching to using the daemon server app.start command
- pass through all logging to the console
- hook up the app termination
- display daemon progress messages as atom jobs
- hook up app restart in the debugger and console UIs

@danrubel 

screen shot 2016-06-10 at 8 20 56 am
screen shot 2016-06-10 at 8 21 03 am

fix https://github.com/flutter/flutter/issues/4249


lib/flutter/flutter_daemon.dart
```diff
@@ -384,6 +408,117 @@ class AppDomain extends Domain {
...
+    }));
+  }
+
+  Future<bool> restart(String appId) => daemon.restart(appId);
+
+  Future<bool> stop(String appId) {
+    return daemon.stop(appId).timeout(
+      new Duration(seconds: 2),
+      onTimeout: () {
+        _dispose();
```
danrubel: Do you want to always call _dispose here? Will there ever be a situation where daemon.stop completes, but onAppStop event is never received thus _dispose will not be called?

lib/flutter/flutter_launch.dart
```diff
@@ -195,55 +179,91 @@ class _RunLaunchInstance extends _LaunchInstance {
...
+      _app = daemon.app.createDaemonApp(result.appId, supportsRestart: result.supportsRestart);
+      _launch.app = _app;
+
+      launchManager.addLaunch(_launch);
+
+      _LogStatusJob job;
+
+      _app.onDebugPort.then((DebugPortAppEvent event) {
+        _observatoryPort = event.port;
+        new Future.delayed(new Duration(milliseconds: 100), _connectToDebugger);
```
danrubel: What happens if you don't wait 100ms ? Add a comment indicating the problem that this delay addresses?

danrubel:
LGTM

lib/flutter/flutter_daemon.dart
devoncarew: Hmm, if we get the `onAppStop`, then _dispose will be called. The timeout here is a guard against the daemon process not being able to stop the app for some reason. It should be rare, but in the event the stop does fail, this timeout will let us remove the app launch from the UI.

lib/flutter/flutter_launch.dart
devoncarew: I'll double check that this is still necessary, and add a comment if so.

Issue #1044, opened 2016-06-22
Uncaught Error: Cannot set property '_wasPauseAtStart' of undefined


tvolkert:
1. Launch Flutter gallery

**Atom Version**: 1.8.0
**System**: Mac OS X 10.11.5
**Thrown From**: [dartlang](https://github.com/dart-atom/dartlang) package, v0.6.27

### Stack Trace

Uncaught Error: Cannot set property '_wasPauseAtStart' of undefined

```
At /Users/tvolkert/.atom/packages/dartlang/web/entry.dart.js:27055

TypeError: Cannot set property '_wasPauseAtStart' of undefined
    at ObservatoryConnection.dart.ObservatoryConnection._handleDebugEvent$1 (/Users/tvolkert/.atom/packages/dartlang/web/entry.dart.js:38452:41)
    at [object Object].<anonymous> (/Users/tvolkert/.atom/packages/dartlang/web/entry.dart.js:28375:32)
    at _RootZone.dart._RootZone.runUnaryGuarded$2 (/Users/tvolkert/.atom/packages/dartlang/web/entry.dart.js:48848:20)
    at _BroadcastSubscription.dart._BufferingStreamSubscription._sendData$1 (/Users/tvolkert/.atom/packages/dartlang/web/entry.dart.js:48096:20)
    at _DelayedData.dart._DelayedData.perform$1 (/Users/tvolkert/.atom/packages/dartlang/web/entry.dart.js:48241:18)
    at _PendingEvents_schedule_closure.dart._PendingEvents_schedule_closure.call$0 (/Users/tvolkert/.atom/packages/dartlang/web/entry.dart.js:48295:16)
    at [object Object].dart._microtaskLoop (/Users/tvolkert/.atom/packages/dartlang/web/entry.dart.js:46046:27)
    at [object Object].dart._startMicrotaskLoop (/Users/tvolkert/.atom/packages/dartlang/web/entry.dart.js:46052:11)
    at TimerImpl_internalCallback0.dart.TimerImpl_internalCallback0.call$0 (/Users/tvolkert/.atom/packages/dartlang/web/entry.dart.js:27433:23)
    at invokeClosure_closure.dart.invokeClosure_closure.call$0 (/Users/tvolkert/.atom/packages/dartlang/web/entry.dart.js:29312:29)
    at _IsolateContext.dart._IsolateContext.eval$1 (/Users/tvolkert/.atom/packages/dartlang/web/entry.dart.js:27046:25)
[...]

PR #1046, opened 2016-06-24
fix a race condition with debugger startup

devoncarew:
- fix a race condition when starting the debugger (using an isolate before it was registered) - fix https://github.com/dart-atom/dartlang/issues/1044
- regen the analysis server protocol

@danrubel


danrubel:
LGTM

PR #1053, opened 2016-06-27
test runner improvements

### LINKED ISSUE #1047: "find references" often doesn't finish
Scenario:
- Select "Find References" on a symbol in a Dart file.
- References sidebar appears
- It shows "searching..."
- No further results ever appear.

I don't see any obvious difference in the logs between a search that succeeds and one that fails.

In the JavaScript console, I see:
 jobs: starting job Find references
 --> search.findElementReferences
jobs: finished job Find references
 <-- a "result" response
 <-- a "search.results" event

Once it stops working, usually (but not always) I have to "reanalyze sources" to get it working again. Then it works once or twice (maybe) and stops working again.

Atom 1.7.4, dartlang 0.6.27
Dart SDK 1.17.0-dev.2.0

### LINKED ISSUE #1050: send client id and version to analysis server
I noticed that IDEA sends these flags when it starts the analysis server:

        "--client-id=IntelliJ_IDEA",
        "--client-version=IU-145.971",

Could the Atom plugin set these flags as well? (It's useful for logging.)

devoncarew:
Several miscellaneous fixes:

- fixed an issue with toString() evaluation in the debugger
- fixed a race condition in the find references feature (fix #1047)
- the test runner will now run tests associated with the current file (running `foo.dart` with the test runner will run the associated `test/foo_test.dart` file)
- better handle ansi escape codes in the console
- added a `create test` command; this creates a new test file corresponding to the current active editor
- send atom client info to the analysis server (fix #1050)

@danrubel


lib/analysis/references.dart
```diff
@@ -49,7 +49,9 @@ class FindReferencesHelper implements Disposable {
         } else {
           bool isMethod = result.element.parameters != null;
           String name = "${result.element.name}${isMethod ? '()' : ''}";
-          FindReferencesView.showView(new ReferencesSearch('References', name, searchId: result.id),
+          Future<List<SearchResult>> resultsFuture = analysisServer.waitForSearchResults(result.id);
```
danrubel: waitForSearchResults --> searchResults?
The "waitFor" implies synchronous to me and this returns with a future.

lib/analysis_server.dart
```diff
@@ -249,7 +250,7 @@ class AnalysisServer implements Disposable {
     if (isActive) _server.analysis.reanalyze();
   }
 
-  Stream<SearchResult> filterSearchResults(String id) {
+  Stream<SearchResult> _filterSearchResults(String id) {
```
danrubel: _filterSearchResults --> _searchResultsStream ?
That way it would align nicely with the rename suggested above.

lib/impl/testing_utils.dart
```diff
@@ -0,0 +1,36 @@
+
+/// Return the list of possible test paths. The given [path] must be in the
+/// `lib/` directory.
+List<String> getPossibleTestPaths(String path, { String separator: '/' }) {
```
danrubel: Perhaps change
```
{ String separator: '/' } {
```
to
```
{ String separator } {
  separator ??= fs.separator;
```
?

That way most (all?) callers would not have to remember to pass in the separator and we're less likely to have a Windows specific bug.

lib/impl/testing.dart
```diff
@@ -83,6 +95,88 @@ class TestManager implements Disposable {
...
+  group('${groupName}', () {
+    test('todo', () {
+      // TODO: Implement test.
+
+    });
+  });
+}
+''');
+
+    await new Future.delayed(Duration.ZERO);
```
danrubel: Please add a comment as to why the delay is needed.

lib/impl/testing.dart
```diff
@@ -83,6 +95,88 @@ class TestManager implements Disposable {
...
+    atom.workspace.open(testPath);
+
+    await new Future.delayed(Duration.ZERO);
```
danrubel: Ditto.

test/testing_utils_test.dart
```diff
@@ -0,0 +1,28 @@
+import 'package:atom_dartlang/impl/testing_utils.dart';
+import 'package:test/test.dart';
+
+main() => defineTests();
+
+defineTests() {
+  group('test_utils', () {
+    test('getPossibleTestPaths', () {
+      expect(getPossibleTestPaths('bin/foo.dart'), equals([]));
```
danrubel: I recommend using join rather than '/' so that tests will use Windows paths on Windows. Alternately create a utility that will accept posix path and return os path so that all tests can be written with posix paths (easier to read than lots of join calls). Perhaps something like...
```
expect(getPossibleTestPaths(p('bin/foo.dart')), equals([]));
```
where `p(String posix)` is a local utility that converts a posix path to an os path.

danrubel:
LGTM with several comments that you can do with as you see fit.

lib/analysis/references.dart
devoncarew: sounds good, will rename

lib/analysis_server.dart
devoncarew: renamed

lib/impl/testing_utils.dart
devoncarew: Perhaps I can make this an assert != null? I'm working around an issue here where this library can't import any of the node wrappers directly, as those import dart:html, and I want this code to be able to be run by CLI unit tests. It's a fair bit of indirection :)

devoncarew: I made the param required, and asserted that it's not null.

lib/impl/testing.dart
devoncarew: done

lib/impl/testing.dart
devoncarew: Updated to wait on the editor being opened from above (so we don't have to have an arbitrary delay).

test/testing_utils_test.dart
devoncarew: Like the `p(String posix)` idea - will do that.

devoncarew: (file https://github.com/dart-atom/dartlang/issues/1055)

lib/impl/testing_utils.dart
danrubel: Ah! I understand the restriction now. Yes, required and asserted makes sense to me.

PR #1054, opened 2016-06-27
optionally enable reload sources when starting an app

devoncarew:
Optionally enable reload sources when starting an app. The flutter devices that support this command will response with a `supportsRestart` property in their app started event.

@danrubel


danrubel:
LGTM

PR #1060, opened 2016-06-29
show flutter app route

devoncarew:
Show the route of the current flutter app in the debugger (fix https://github.com/dart-atom/dartlang/issues/954).

@danrubel


danrubel:
LGTM

PR #1062, opened 2016-07-01
also support tests in test/all.dart

devoncarew:
- remove the support for showing flutter app routes
- also support running tests in files called `test/all.dart`

@danrubel


lib/impl/testing.dart
```diff
@@ -45,7 +47,9 @@ class TestManager implements Disposable {
     path = _findAssociatedTest(project, path);
 
     if (!allowWithoutTestName) {
-      if (!path.endsWith('_test.dart')) return false;
+      if (!path.endsWith('_test.dart') && !path.endsWith('${_sep}test${_sep}all.dart')) {
```
danrubel: To support `all.dart` in child `test` directories rather than just the top level `test` directory...
```
(!path.contains('${_sep}test${_sep}') || !path.endsWith(${_sep}all.dart'))

danrubel:
LGTM!

lib/impl/testing.dart
devoncarew: :+1:

PR #1066, opened 2016-07-14
fix a websocket origin issue

devoncarew:
- fix an issue with debug protocol changes - use a node WebSocket implementation instead of the browser based one (it was reporting `file://` as the socket origin)
- improve an error message when launching Flutter apps
- allow more than one Flutter app to run at a time
- when re-running an already running app, choose to restart the app if the target supports it

@danrubel


lib/flutter/flutter_launch.dart
```diff
@@ -68,7 +70,27 @@ class FlutterLaunchType extends LaunchType {
         "settings for the 'flutter' plugin and / or try re-starting Atom.";
     }
 
-    await _killLastLaunch();
+    // Instead of killing the last launch, check if we should re-start it.
+    if (_lastLaunch != null) {
```
danrubel: Combine the two `if (_lastLaunch != null) {` statements ?

lib/debug/websocket.dart
```diff
@@ -0,0 +1,56 @@
+import 'dart:async';
+import 'dart:js';
+
+import 'package:atom/node/node.dart';
+
+class WebSocket {
```
danrubel: At some point in the future, should we move this class into the `atom.dart` plugin?

danrubel:
LGTM

lib/debug/websocket.dart
devoncarew: Yup, will add a TODO:

lib/flutter/flutter_launch.dart
devoncarew: done

devoncarew:
Thanks for the review! Landing...

PR #1074, opened 2016-08-10
save editors before issuing an app restart command

devoncarew:
- save editors before issuing an app restart command

@danrubel


danrubel:
LGTM

PR #1075, opened 2016-08-17
add the ability to send in a full application restart

devoncarew:
- add the ability to send in a full application restart
- this works via shift-click on the restart button, or cmd-shift-r (instead of cmd-r, for run/restart)

@danrubel


danrubel:
LGTM

PR #1076, opened 2016-08-19
re-apply futter debug options between full app restarts

devoncarew:
- re-apply flutter debug options between full app restarts

@danrubel


danrubel:
LGTM

PR #1083, opened 2016-09-12
change the 'enable hot reload' setting to a 'disable hot reload' one

devoncarew:
- change the 'enable hot reload' setting to a 'disable hot reload' one (we now rely on the default setting in flutter_tools)
- update the version of the analysis server library
- fix one strong mode issue

@danrubel


danrubel:
LGTM

PR #1092, opened 2016-10-12
Fix syntax highlighting of TODO, FIXME, etc

tejainece:


tejainece:
I don't know why but `match` prevents `TODO` syntax highlighting. When `begin` and `end` are used, it works.

The `match` doesn't do nothing special anyway in this case.

devoncarew:
Can you attach a screen shot of the issue? I see `// TODO`s being highlighted correctly.

devoncarew:
Closing; feel free to re-open if you think this is valid.

Issue #1093, opened 2016-10-13
Cannot create breakpoints


KOGI:
I am unable to create breakpoints anywhere in my code. Every time, I'm greeted with "Uncaught Error: Cannot decorate a destroyed marker".

Here's the stack trace:

[Enter steps to reproduce below:]

1. Open dart project in Atom
2. Open any `.dart` file
3. Click in the gutter to add a breakpoint (or press F9)

**Atom Version**: 1.11.1
**Electron Version**: 0.37.8
**System**: Microsoft Windows 10 Enterprise
**Thrown From**: [dartlang](https://github.com/dart-atom/dartlang) package, v0.6.40

### Stack Trace

Uncaught Error: Cannot decorate a destroyed marker

```
At C:\Users\Bob\.atom\packages\dartlang\web\entry.dart.js:28377

Error: Cannot decorate a destroyed marker
    at DecorationManager.module.exports.DecorationManager.decorateMarker (C:\Users\Bob\AppData\Local\atom\app-1.11.1\resources\app.asar\src\decoration-manager.js:192:15)
    at TextEditor.module.exports.TextEditor.decorateMarker (C:\Users\Bob\AppData\Local\atom\app-1.11.1\resources\app.asar\src\text-editor.js:1657:37)
    at JsObject.dart.JsObject.callMethod$2 (C:\Users\Bob\.atom\packages\dartlang\web\entry.dart.js:46228:44)
    at TextEditor.dart.ProxyHolder.invoke$4 (C:\Users\Bob\.atom\packages\dartlang\web\entry.dart.js:49146:27)
    at TextEditor.dart.ProxyHolder.invoke$3 (C:\Users\Bob\.atom\packages\dartlang\web\entry.dart.js:49162:21)
    at TextEditor.dart.TextEditor.decorateMarker$2 (C:\Users\Bob\.atom\packages\dartlang\web\entry.dart.js:48998:46)
    at _EditorBreakpoint.dart._EditorBreakpoint._EditorBreakpoint$4 (C:\Users\Bob\.atom\packages\dartlang\web\entry.dart.js:56379:21)
    at undefined.dart._EditorBreakpoint.static._EditorBreakpoint$ (C:\Users\Bob\.atom\packages\dartlang\web\entry.dart.js:56386:14)
    at BreakpointManager.dart.BreakpointManager._createEditorBreakpoint$2 (C:\Users\Bob\.atom\packages\dartlang\web\entry.dart.js:56087:40)
[...]

PR #1094, opened 2016-10-19
fix an issue with breakpoints

devoncarew:
- fix https://github.com/dart-atom/dartlang/issues/1093
- we were not tracking the list of open editors correctly; this caused an issue when we tried to set a breakpoint in a disposed editor


KOGI:
Thank you!

PR #1106, opened 2016-11-18
fix #1105

### LINKED ISSUE #1105: Hotkey for hiding dartdoc
I want open and hide dartdoc popup with keyboard (i use keyboard only and cant use tooltips).

averrin:
add toggling control into _handleDartdoc


devoncarew:
lgtm!

PR #1130, opened 2017-07-26
Patches for atom 1.18

cdavidjean:
Fixed startup for atom 1.18
Fixed pub calls
Fixed linter/squiggles


lib/impl/tooltip.dart
```diff
@@ -116,19 +131,24 @@ class TooltipElement extends CoreElement {
       : super('div', classes: 'hover-tooltip') {
     id = 'hover-tooltip';
 
+    print(position);
```
devoncarew: We should remove this.

package.json
```diff
@@ -26,9 +26,9 @@
         "^1.0.0": "consumeToolbar"
       }
     },
-    "linter-plus-self": {
+    "linter-indie": {
```
devoncarew: Do we need to update the required-packages list above as well?

devoncarew:
I landed the atom.dart plugin, so the travis build job now finishes cleanly.

package.json
Arcanemagus: Depends on whether whatever is consuming the `required-packages` key supports version numbers as that service is only available with `linter` v2.

cdavidjean:
Oh damn...  like I said I'm new to git, i thought the pull request would
use a snapshot copy...  i am not done with reviving the tootips yet.
Should i try to uncommit those changes?

devoncarew:
> Should i try to uncommit those changes?

I would either revert those changes and update the PR, or keeping working and I can re-review when you're done. You'll have a bit of a git dance to do, so the 2nd option would likely be easier for you (given that you're new to git. Happy either way :)

cdavidjean:
Thanks, I'll take option two :). It shouldn't be too long, probably some
time tomorrow.

lib/impl/tooltip.dart
cdavidjean: Done.

cdavidjean:
A Better summary of this PR (which is now ready for review, sorry):
- Fixed jumpKey navigation
- Fixed linter service (using newer linter api)
- Fixed outline by attaching to editor element instead of shadowDom root.
- Fixed some startup NPE.
- Revived Tooltips, moved them above (not to clash with linter) and added debouncing.
- Added tracking subscription for TooltipManager.

devoncarew:
Awesome! Thanks much, landing.

Issue #1118, opened 2017-03-15
Lints no longer show up in IDE


jacehensley-wf:
After upgrading to [Linter v2](http://steelbrain.me/2017/03/13/linter-v2-released.html) infos, warnings, and errors are not showing up in the IDE. They still show up in the warning panel.

![image](https://cloud.githubusercontent.com/assets/12398382/23971199/32b02346-098a-11e7-9964-cebd855d8ac9.png)

devoncarew:
From looking over that blog post, it's likely that we have some API changes to deal with to work with the v2 of the linter -

jacehensley-wf:
I'd be happy to help out if you point me in some general direction 😄

devoncarew:
Great!

This file: https://github.com/dart-atom/dartlang/blob/master/lib/atom_linter.dart is our interface to the linter API. I suspect (but haven't checked) that this is where things may have broken.

We define our our UI for how to show lints (the errors view). It looks like the v2 linter has both a better UI than the v1 linter and a configurable one. So - a much more comprehensive bit of work - would probably involve removing our custom problems view. But just getting things back to working would be a good start :)

PR #1135, opened 2017-08-03
Removed dart error list view in favor of linter view

cdavidjean:
- Moved to linter-indie v2.
- Removed dart error list view in favor of linter view.
- Moved Fixit handling through linter (ALT-ENTER).  For now all we pass to linter is a
  'Quick Fix for ...' entry for each place it is available.  When one is 'applied',
  we use the analyzer to aync compute the actual list of possible fixes.  I'll
  make a change request to linter to allow passing in a Promise to create the 
  list of solutions, then we can get rid of the intermediate step.
- Added passing of dart analyzer suggested correction to the linter.
- Added Fixit icon to outline error list
- Fixed some tooltips focus issue, by making Debouncer cancelleable.

fix https://github.com/dart-atom/dartlang/issues/1118


lib/linter.dart
```diff
@@ -21,8 +22,28 @@ LintMessage _errorToLintMessage(String filePath, AnalysisError error) {
...
+  // hint button that will compute and open all fixes.
+  return [new LintSolution(
+      title: 'Quick Fixes for ${error.message}',
+      position: _locationToRange(error.location),
+      apply: showQuickFixes)];
+}
+
+void showQuickFixes() {
+  // Show the quick fix menu.
+  print('QUICK FIX');
```
devoncarew: You can probably remove this debug output.

lib/atom_linter.dart
```diff
@@ -66,7 +67,26 @@ class LinterService extends ProxyHolder {
...
+    _linter.invoke('setAllMessages', list);
+  }
+}
+
+class LintSolution {
+
+  final String title;
+  final num priority;
+  final Rn position;
+  Function apply;
```
devoncarew: Perhaps use a typedef instead of `Function`? Something like: `typedef void VoidFunc();`

lib/impl/errors.dart
```diff
@@ -1,386 +0,0 @@
-
-/// A view and status line decoration for visualizing errors.
-library atom.errors;
```
devoncarew: ah, but it was such a nice UI

lib/plugin.dart
```diff
@@ -331,7 +319,7 @@ class AtomDartPackage extends AtomPackage {
       // show infos and todos
       'configureErrorsView': {
         'title': "Errors view configuration",
-        'description': 'Choose which types of items to show in the errors view.',
+        'description': 'Choose which types of items to show in the Linter.',
```
devoncarew: Is it generally referred to as the `Linter`? `linter view`?

Arcanemagus: That's a bit tricky to answer, `linter` is "just" the package that manages all the providers and processes their results into a unified set. Built on top of that are UIs that display those results to users, `linter-ui-default` being the bundled one.

Personally I'd probably just go with "... show in Linter" (ie. remove the `the`), since the full UIs out there display themselves as that.

lib/impl/errors.dart
cdavidjean: At least I moved the quick fix icon to the outline view :)

lib/plugin.dart
cdavidjean: I went with the smallest one 'Linter'.

lib/plugin.dart
```diff
@@ -331,7 +319,7 @@ class AtomDartPackage extends AtomPackage {
       // show infos and todos
       'configureErrorsView': {
         'title': "Errors view configuration",
-        'description': 'Choose which types of items to show in the errors view.',
+        'description': 'Choose which types of items to show in Linter.',
```
devoncarew: Honestly, this doesn't read quite right to me; `in the linter view.` I think would read better, eps. in the context of the other settings. However, we can land this and iterate.

devoncarew:
Thanks for the PR! Great to have the plugin modernized.

PR #1140, opened 2017-08-09
Added pub serve from IDE

cdavidjean:
- Added pub serve to context menus
- Added pub serve launch, so we can track process from a console window

(sorry for the noise in the commits list?)


lib/launch/launch_serve.dart
```diff
@@ -0,0 +1,69 @@
...
+
+/// Pub serve launch.
+class ServeLaunchType extends LaunchType {
+  static void register(LaunchManager manager) =>
+      manager.registerLaunchType(new ServeLaunchType());
+
+  ServeLaunchType() : super('serve');
+
+  bool canLaunch(String path, LaunchData data) {
+    return path.endsWith('.yaml');
```
devoncarew: Why `.yaml`? Do you mean pubspec.yaml?

lib/impl/pub.dart
```diff
@@ -74,6 +84,7 @@ class PubManager implements Disposable, ContextMenuContributor {
 
   List<ContextMenuItem> getTreeViewContributions() {
     return [
+      new PubContextCommand('Pub Serve', 'dartlang:pub-serve', true),
```
cdavidjean: Done.  My thought was that Pub Serve is run more often than Pug Get.

cdavidjean:
Done, thanks

PR #1142, opened 2017-08-12
rev to 0.6.49

devoncarew:
@cdavidjean - this updates the version to `0.6.49` in order to release the recent changes:

- removed the Dart errors view in favor of using the Linter's default UI
- some fixes to the outline view's styles
- updated view management to use atom's docks
- fixed rename

given the scope of the changes, we may instead want to rev to `0.7.0`; thoughts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # dartlang plugin changelog
 
-## 0.6.49
+## 0.7.0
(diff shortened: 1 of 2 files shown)


cdavidjean:
I think 0.7.0 sounds good.  We have fixes and one new features.

devoncarew:
👍 - bumped to `0.7.0`

PR #1143, opened 2017-08-15
Added web launch

cdavidjean:
- Added web launch
- Fixed some pub serve launch issues
- Fixed some console related issues introduced by the new DockViews.


cdavidjean:
Note that: https://github.com/dart-atom/atom.dart/pull/52 has to be merged in first.

lib/launch/launch_web.dart
```diff
@@ -13,9 +24,54 @@ class WebLaunchType extends LaunchType {
...
-    // TODO:
-    return new Future.error(new UnimplementedError());
+    Browser browser = deps[BrowserManager].browser;
+    if (browser == null) {
+      atom.notifications.addWarning('No browser configured.');
+      return new Future.value();
+    }
+
+    // Find pub serve for 'me'.
+    ServeLaunch pubServe = manager.launches.firstWhere((l) =>
```
devoncarew: Another option is to always start and stop a pub serve instance for a launch - that the `pub serve` instance has the same lifetime as the launch.

lib/plugin.dart
```diff
@@ -299,6 +302,15 @@ class AtomDartPackage extends AtomPackage {
         'order': 1
       },
 
+      // debugger
+      'debugBrowserLocation': {
```
devoncarew: Perhaps just `browserLocation`?

lib/plugin.dart
```diff
@@ -299,6 +302,15 @@ class AtomDartPackage extends AtomPackage {
...
+        'title': 'Debug Browser',
```
devoncarew: Chrome Location? Browser Location?

lib/plugin.dart
```diff
@@ -299,6 +302,15 @@ class AtomDartPackage extends AtomPackage {
...
+        'description': 'The location of a Browser with Chrome DevTools Protocol.',
```
devoncarew: You may just want to call out Chrome specifically.

lib/browser.dart
```diff
@@ -0,0 +1,91 @@
...
+
+    return runner.onExit.then((_) => buf.toString());
+  }
+
+  List<String> execArgsFromYaml(yamlArgs) {
+    List execArgs = [];
+    if (yamlArgs is Map) {
+      yamlArgs.forEach((k, v) {
+        if (singletonBoolParameters.contains(k)) {
+          if (v) execArgs.add('--$k');
```
devoncarew: do we need to check here if `v` is a `bool`?

lib/launch/launch_web.dart
devoncarew: In that setup, the `serve` output would be just part of the launches output, and user's wouldn't have to know to start the `pub serve` before the web launch.

cdavidjean: You could also debug more than one html from the same project, having one pub serve for that use case is nice.
I'll add that as an option once the debugger works.  For now it's easier to build small steps.

lib/plugin.dart
cdavidjean: Done.

lib/plugin.dart
cdavidjean: Done.

lib/plugin.dart
cdavidjean: The debug protocol seems to be used by more than one browser, according to:
https://github.com/cyrus-and/chrome-remote-interface#implementations
I'm happy either way.

lib/browser.dart
cdavidjean: AFAICS the Yaml parser seems to turn true and false into boolean.

lib/plugin.dart
devoncarew: Hmm, perhaps something like:

```
The path to Chrome (or other browser supporting the Chrome DevTools protocol).
```

cdavidjean: Perfect.  Done.

PR #1147, opened 2017-08-25
First useable web debugger launch

cdavidjean:
Sorry about huge pr, I think this is mostly an all or nothing situation.


devoncarew:
Sorry for the delay! WIll take a look at this this evening.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # dartlang plugin changelog
 
+## unreleased
+- added browser selection
+- added html launch
+- added web debugger from identifying
+- fixes to console window and pub servce launch
```
devoncarew: sp: service

lib/browser.dart
```diff
@@ -75,10 +75,11 @@ class Browser {
     return runner.onExit.then((_) => buf.toString());
   }
 
-  List<String> execArgsFromYaml(yamlArgs) {
+  List<String> execArgsFromYaml(yamlArgs, {List<String>exceptKeys: const []}) {
```
devoncarew: nit: missing a space (I would suggest just dartfmt the file)

lib/debug/chrome.dart
```diff
@@ -0,0 +1,790 @@
...
+  ///
+  /// [lineNumber]: Line number to set breakpoint at.
+  /// [url]: URL of the resources to set breakpoint on.
+  /// [urlRegex]: Regex pattern for the URLs of the resources to set breakpoints
+  ///   on. Either url or urlRegex must be specified.
+  /// [columnNumber]: Offset in the line to set breakpoint at.
+  /// [condition]: Expression to use as a breakpoint condition. When specified,
+  ///   debugger will only stop on the breakpoint if this expression evaluates
+  ///   to true.
+  Future<Breakpoint> setBreakpointByUrl(int lineNumber,
```
devoncarew: Just FYI, in a previous dart web debugger I wrote, we used setBreakpointByRegex. It meant we were less reliant on the url prefix, and on the specific path that the user found to run code from a `packages/` subdir.

lib/debug/chrome_debugger.dart
```diff
@@ -0,0 +1,638 @@
...
+    if (isPaused) {
+      isPaused = false;
+      chrome.debugger.stepOver();
+    }
+  }
+
+  autoStepOver() => stepOver();
+
+  // TODO: do we need this?
+  stepOverAsyncSuspension() {}
```
devoncarew: I don't think this would have an analog when debugging web apps (so, a no-op impl seems fine).

lib/debug/model.dart
```diff
@@ -33,9 +34,11 @@ abstract class DebugConnection {
 // TODO: Add an IsolateState class.
 
 /// A representation of a VM Isolate.
-abstract class DebugIsolate {
+abstract class DebugIsolate extends MItem {
```
devoncarew: I haven't looked closely into where you're using these now, but this file previously had a separation of concerns between non-ui (model) and UI code. I think it's still likely worth keeping this separation.

lib/debug/observatory_debugger.dart
```diff
@@ -468,14 +469,18 @@ String _refToString(dynamic value) {
 
 Point _calcPos(Script script, int tokenPos) {
   List<List<int>> table = script.tokenPosTable;
+  if (table == null) return null;
 
   for (List<int> row in table) {
-    int line = row[0];
-
+    if (row == null || row.isEmpty) continue;
+    //int line = row[0]; <- this crashed, don't know why
```
devoncarew: you can remove the commented out code

lib/material.dart
```diff
@@ -103,17 +103,28 @@ typedef void ListRenderer(dynamic obj, CoreElement element);
 typedef bool ListFilter(dynamic obj);
 typedef int ListSort<T>(T obj1, T obj2);
 
+abstract class MItem {
```
devoncarew: Ah, ok, an MItem just provides some labelling info (not access to the DOM). Perhaps move out of the material library, so we don't need a ref to material from the model library?

lib/debug/chrome.dart
cdavidjean: Can you give me an example of how you built the regex?

devoncarew: It would be from memory from a few years ago, but I believe that:
- anything from a package (package: or a self-ref to the lib/ for the current project) was something like `.*/packages/foo_package/foo.dart`
- in a web/ dir (or example/, or test/), we made the regex relative to that dir (`.*/my_app.dart` from /web/my_app.dart)

cdavidjean:
Done.  Thanks!

PR #1149, opened 2017-08-31
Fixes to debugger

cdavidjean:
- Fetching and merging own properties and inherited ones.
- Fixed crash on deselecting item from MTree


lib/debug/chrome.dart
```diff
@@ -728,13 +719,13 @@ class PropertyDescriptor extends InternalPropertyDescriptor {
   String toString([String indent = '  ']) =>
       "PD: $name ${value?.toString(indent + '  ')}\n"
       "${indent}fg: w: $writable c:$configurable e:$enumerable t:$wasThrown o:$isOwn\n"
-      "${indent}get: $getFunction\n"
-      "${indent}set: $setFunction\n"
+      "${indent}get: ${getFunction.toString(indent + '  ')}\n"
```
devoncarew: Hmm, not necessarily part of this PR, but it's odd to see `toString()` with an additional param. This is to map 1:1 to the with the JS chrome library?

cdavidjean: It's a trick I use to allow formatting toString output but since it covers the regular toString() then you can use it in regular print or string formatting.

PR #1150, opened 2017-09-04
Added ddc unparsing option of stack/scope for better UX

cdavidjean:
- combined scopes for local function
- rename __proto__ to super
- sort super to the end
- allow dynamic evaluation of getters
- remove Symbols for functions
- rename Symbol(a.b.c) to c, allowing merging with (likely) getter c also present on 'this'.
- added async handling (this + async stack frame)


devoncarew:
lgtm!

PR #1151, opened 2017-09-04
Fixes on debugger for Windows 

cdavidjean:


cdavidjean:
Devon,
I still need to fix breakpoints in packages, using regex as you suggested.  If possible, please wait until you build a new release.
Thanks,
David

cdavidjean:
I fixed and refactored breakpoints and it now works over packages.
I seem to have an issue with tooltips on either Windows or Atom 1.9, but I don't think it is not blocking for now.  Thanks for your patience over these long pr.

lib/debug/chrome_debugger.dart
```diff
@@ -189,77 +209,100 @@ class ChromeConnection extends DebugConnection {
     }).catchError((e) {
       launch.pipeStdio('Error setting async call depth ($e).\n');
     }).then((_) {
+      return installBreakpoints();
```
devoncarew: I notice that you're not using `async` / `await` for implementing method like this. It can make them a bit terser, and really simplifies error handling. The above might look like:

```
try {
  await chrome.debugger.setPauseOnExceptions(
       exceptionTypes[breakpointManager.breakOnExceptionType]);
} catch {
  launch.pipeStdio('Error setting exception mode ($e).\n');
  return;
}

try {
  await chrome.debugger.setAsyncCallStackDepth(16);
} catch {
  launch.pipeStdio('Error setting async call depth ($e).\n');
  return;
}

await installBreakpoints();
return chrome.page.navigate(url);
```

cdavidjean: I was under the impression that the plugin didn't use async/await, but I guess it's because most of its code pre-dates async/await.  
Fixed most uses in the file.  Much more readable.
I love dart!

cdavidjean:
While fixing the windows issue I discovered that dart getters are not handled well in the scope list, I'll fix that tomorrow.

lib/debug/chrome_debugger.dart
devoncarew: Yeah, a lot of it predates async await being available (and me being familiar enough with it). I still use explicit Future and Completers for some specific use case, but async await for the majority of things.

PR #1152, opened 2017-09-09
Added Library tab for web debug

cdavidjean:
- Added async updating of library tab
- Added hack to be able to open packages file for package not in analyzer root.


lib/debug/debugger.dart
```diff
@@ -160,8 +163,26 @@ class UriResolver implements Disposable {
     if (_uriToPath.containsKey(uri)) return _uriToPath[uri];
 
     String contextId = await _completer.future;
-    MapUriResult result = await analysisServer.server.execution.mapUri(contextId, uri: uri);
-    String path = result.file;
+    String path;
+    try {
```
devoncarew: Hmm, I'm curious what this is doing. You're using the analysis server to resolve a uri to a file path? There is an API for that, in the debugger domain.

I may not understand what this is trying to do though - I'd suggest adding some descriptive comments :)

devoncarew: It is at least very creative :)

lib/debug/debugger_ui.dart
```diff
@@ -120,6 +121,9 @@ class DebuggerView extends DockedView {
     subs.add(connection.onPaused.listen(_handleIsolatePaused));
     subs.add(connection.onResumed.listen(_handleIsolateResumed));
     subs.add(connection.isolates.onRemoved.listen(_handleIsolateTerminated));
+
+    // Optional events
+    subs.add(connection.onLibrariesChanged?.listen(_handleLibrariesChanged));
```
devoncarew: Ah, this will add a `null` subscription sometimes. You may instead want to re-write to use an if statement.

cdavidjean: Done.

lib/debug/debugger.dart
cdavidjean: :)  Added comment.  If you think it is an analyzer bug, please let me know.

PR #1153, opened 2017-09-10
use package:analysis_server_lib

devoncarew:
- switch to using then internal generated analysis_server_lib.dart to `package: analysis_server_lib`. This is a largely mechanical change; `package: analysis_server_lib` was originally copied from this project, and has evolved a bit from there. Using this package makes this project analyze a bit more cleanly with the latest strong mode changes.

@cdavidjean


cdavidjean:
Great cleanup, thanks!

PR #1154, opened 2017-09-10
Dockable outline view

cdavidjean:
The outline panel is breaking more and more with updates to atom.  In 1.19.5, for example:
- resizing outline hides the editor scrollbar 
- mousewheel scrolling does it on outline and the editor underneath.
This moves it into a (single) dockable panel that the user can move around like any other panel.  Feel free to veto it out :)


devoncarew:
FYI, there are a fair number of other commits in here (including ones that look like they're already merged). This will all clean up via the 'squash and merge' button, but you may want to double check your git workflow.

cdavidjean:
Even with a branch I find it hard to sync correctly.  I think I finally figured if out tonight, for my next fix...
On the bright side, the 4 files list as changed are the correct changes, two of them being the .js and .map files.

devoncarew:
One other regression I noticed with the current outline is that resizing the splitter doesn't track the mouse well.

I agree that the fixes to the above issues are an improvement. I do like the current (in-line in the editor) style, but don't have cycles to fix it up. So, lgtm for this PR :)

One thing we get from the current solution is that each editor has its own state for the outline view (selection, scroll position). After moving to a shared outline view per window (instead of per editor), we may want to tweak the behavior a bit so that switching form editor A to B and back to A, the outline view's state for editor A is preserved. If this makes sense, we could track the work in an issue.

cdavidjean:
Added issue 1156 for tracking outline caching.  Thanks!

PR #1155, opened 2017-09-10
fix strong mode issues

devoncarew:
- fix ~14 strong mode issues w/ the latest sdk (using `dynamic` as a bottom type)

@cdavidjean

### DIFF
--- a/lib/debug/debugger_ui.dart
+++ b/lib/debug/debugger_ui.dart
@@ -681,7 +681,7 @@ class LibrariesTab extends MTab {
   }
 
   void _render(dynamic _lib, CoreElement element) {
-    ObservatoryLibrary lib = _lib;
+    DebugLibrary lib = _lib;
     element..add([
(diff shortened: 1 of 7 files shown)


lib/debug/debugger_ui.dart
```diff
@@ -678,7 +680,8 @@ class LibrariesTab extends MTab {
     });
   }
 
-  void _render(ObservatoryLibrary lib, CoreElement element) {
+  void _render(dynamic _lib, CoreElement element) {
+    ObservatoryLibrary lib = _lib;
```
cdavidjean: My mistake, but all ObservatoryLibrary here should be DebugLibrary, can you fix?

cdavidjean:
lgtm

lib/debug/debugger_ui.dart
devoncarew: sure!

PR #1166, opened 2017-12-31
Don't show outline view when setting is disabled

### LINKED ISSUE #1164: Disabling "Show outline view" doesn't work
I disabled "Show outline view" because it always opens above the project pane whenever I open atom.
y9po8zib8s

But this setting has no effect. The outline pane still opens all the time.

Atom 1.22.1, atom-dart 1.0.1, SDK 2.0.0-dev.7.0 running on macos.

passsy:
The outline view was opening at startup regardless of the setting "Show outline view". It opened at every atom startup. I had to disable the `dart` plugin for the times I'm not working with dart because it's so annoying.

fixes #1164

### DIFF
--- a/lib/impl/outline.dart
+++ b/lib/impl/outline.dart
@@ -36,7 +36,9 @@ class OutlineController extends DockedViewManager<OutlineView> {
     );
 
     disposables.add(atom.config.observe(_keyPath, null, (val) {
-      showView();
+      if (val == true) {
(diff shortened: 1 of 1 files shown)


devoncarew:
Thanks!

chenmingxiang110:
The outline view is still opening regardless of the "show outline view" setting.

PR #1173, opened 2018-09-05
add the mixin keyword

devoncarew:
- add the `mixin` keyword
- work towards supporting https://github.com/dart-lang/language/issues/12

cc @DanTup


DanTup:
Done the same, thanks!

PR #1179, opened 2018-12-19
support nested comments

devoncarew:
- support nested comments
- fix https://github.com/dart-atom/dart/issues/1082

@DanTup, looks good?


devoncarew:
(we can ignore the build failure for now)

DanTup:
Yep, LGTM (in that it's the same as my fix, but I don't have atom set up to test - it certainly worked in VS Code, and all the related code looks the same too :-))

PR #1194, opened 2019-09-05
Add "extension" keyword to gramma

DanTup:
Related to https://github.com/dart-atom/dart/issues/1193

This is the same change made in VS Code:

https://github.com/Dart-Code/Dart-Code/commit/06eaf66e9613b16e1401dd267c5597dff1524239

Note: I don't know how to run this locally to test, I made the change blind since it seems simple and mirrors VS Code's (which I did test) :-)

### DIFF
--- a/grammars/dart.cson
+++ b/grammars/dart.cson
@@ -206,7 +206,7 @@
         'name': 'keyword.control.new.dart'
       }
       {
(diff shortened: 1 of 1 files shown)


DanTup:
Travis is not happy, but I don't think it's related to this change 😁

> Because atom_dart requires SDK version >=1.9.0 <2.0.0, version solving failed.
> The command "pub get" failed and exited with 1 during .

